PR #112, opened 2011-12-23
Runebar

Pondidum:
Hi Haste,

Finally getting around to submitting this to you!

I reworked the DK runes to work in the same manner as DocsDebugRunes - roughly speaking, the runes work more like one progress bar.

If you have r1 ready, and r2 is filling, you have r1 on the left, and r2 on the right.  if r2 is ready, and r1 is charging, they get swapped so that the ready rune is on the left side of the display.  You will never end up with a depleted rune on the left, and a charging rune on the right.

A couple of screenshots to help demonstrate:
https://lh3.googleusercontent.com/-aMDgT0ZgiUM/TvSbZtm0dmI/AAAAAAAAAtw/z4KAmUKyteo/s0/1.jpg
https://lh3.googleusercontent.com/-ZzQ2caw0jA8/TvSbZoncemI/AAAAAAAAAtw/xomsTvtXkIQ/s0/2.jpg

Not sure if this is suitable for inclusion by default - but if other DKs are anything like me, I cannot play to the best without this logic!

Thanks.


p3lim:
The rune system was simplified in 7.0 which #269 covers, this PR should be irrelevant now.

PR #223, opened 2014-06-25
warlockpowerbar: add the element

Rainrider:
Almost 2 years later :smile: I don't know if anyone would switch to that after this long, but let's give it a try.

It's rather generic in that it only takes care about setting the bar values and toggling their display based on spec and vehicle. It will also change the color of the burning embers based on whether the player has the green fire spell (that is probably a cpu waste for something that cosmetic). It won't re-size the bars when it switches from demonology (single bar) to destruction (4 bars), so the layout will have to take care about that in :PostVisibility() if they wish so.

Would be happy about feedback

### DIFF
--- a/elements/warlockpowerbar.lua
+++ b/elements/warlockpowerbar.lua
@@ -4,12 +4,17 @@
 
  Widget
 
- WarlockPowerBar - an array consisting of 4 UI StatusBars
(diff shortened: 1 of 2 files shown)


Rainrider:
This will be made obsolete by #266

PR #238, opened 2014-11-01
resurrect: Update properly when the unit is successfully resurrected

p3lim:
INCOMING_RESURRECT_CHANGED is not dependable, and I've often seen that once the unit is successfully resurrected, the icon gets stuck.


p3lim:
Seems to have been resolved in Legion.

PR #240, opened 2014-11-20
Since WoW 6.0 GetComboPoints() may return 0 instead of the correct combo point value

zorker:
...0 from time to time

Changed the element to work with UnitPower(unit,4) which was introduced with patch 6.0.
The POWER_TYPE_INDEX of 4 references COMBO_POINTS since 6.0.

### DIFF
--- a/elements/cpoints.lua
+++ b/elements/cpoints.lua
@@ -36,11 +36,12 @@
 local parent, ns = ...
 local oUF = ns.oUF
 
-local GetComboPoints = GetComboPoints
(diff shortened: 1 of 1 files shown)


p3lim:
"Fixed" by #266, as the cpoints element will be deprecated in favor of the classicons element.

p3lim:
UnitPower has the same issue, and Blizzard is aware of this.
http://www.wowinterface.com/forums/showthread.php?t=53496

(Just adding this comment for related info)

Rainrider:
Closing this at there is nothing new. CPoints is going bye bye

PR #241, opened 2014-11-26
Extra units support for :Tag()

p3lim:
This PR contains the following changes:
- 582d8dfe153e4cf8d4195cf05a2a55bef3f71454 Minor change to avoid doing the same check twice
- 2ffcfbc12e516ede3b2b7d5ba5b7860cfc5ffaef This commit allows the layout to specify additional units which the event handler can match against. This is to make sure you could get an update for a tag on one unit for a different unit. As an example; I have my pet's health on my player frame, as I don't see a point of having a whole frame for it. With this I can tag like this: `frame:Tag(fs, '[pethp][othertag]', 'pet')` instead of forcing `UNIT_HEALTH_FREQUENT` or other events to be "unitless" with [oUF.Tags.SharedEvents](https://github.com/p3lim-wow/oUF/blob/1.6.6/elements/tags.lua#L697)

### DIFF
--- a/elements/tags.lua
+++ b/elements/tags.lua
@@ -423,7 +423,7 @@ frame:SetScript('OnEvent', function(self, event, unit)
 	local strings = events[event]
 	if(strings) then
 		for k, fontstring in next, strings do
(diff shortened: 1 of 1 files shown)


p3lim:
Should be fine to merge, I've been using this myself for 2+ years without any issues.

PR #248, opened 2015-07-04
Bump TOC interface version to 6.2

renstrom:


haste:
This is handled by my release script, so not really necessary.

PR #251, opened 2015-07-28
runebar: Bail if GetRuneCooldown returns nil values

p3lim:
Since 6.2, `GetRuneCooldown` returns nil values during loading screens, which result in errors being spewed.

It is most likely an issue with the API, but it's probably best to check for it to avoid errors.

Discussion: http://www.wowinterface.com/forums/showthread.php?t=52535

### DIFF
--- a/elements/runebar.lua
+++ b/elements/runebar.lua
@@ -92,6 +92,11 @@ local UpdateRune = function(self, event, rid)
 	end
 
 	local start, duration, runeReady = GetRuneCooldown(rid)
+	if(not start) then
+		-- As of 6.2.0 GetRuneCooldown returns nil values when zoning
+		return
+	end
+
 	if(runeReady) then
 		rune:SetMinMaxValues(0, 1)
 		rune:SetValue(1)


Pondidum:
I also had to do a similar fix in ouf_Dark for my custom rune bar.  You can emulate the original return values by wrapping calls to `GetRuneCooldown`:

	local getRuneCooldown = function(rid)
		local currentStart, currentDuration, currentReady = GetRuneCooldown(rid)

		return currentStart or 0, currentDuration or 0, currentReady
	end

Also worth noting that `GetRuneType` can also break (might get hit here: [runebar.lua](https://github.com/p3lim-wow/oUF/blob/FixRunebar/elements/runebar.lua#L68)), and return nil rather than the range it should do (1 -> 4), again wrapping the call to restore original functionality:

	local getRuneType = function(rid)
		return GetRuneType(rid) or 1
	end

Rainrider:
@Pondidum Your approach adds an extra function call for no real benefit. Both https://github.com/p3lim-wow/oUF/blob/FixRunebar/elements/runebar.lua#L68 and p3lim's solution are better and sufficient imho.

sdkyron:
Thanks for the fix

PR #252, opened 2015-07-29
altpowerbar: Some tweaks

ls-:
Since 6.2, I guess, Blizzard changed ```UnitAlternatePowerInfo()``` by adding one more return value.

It is also illogical to override possibly existing ```OnLeave``` script.


Rainrider:
@haste could you please merge this

PR #256, opened 2015-09-30
Fix for changes in return values from UnitAlternatePowerInfo

Phanx:
It looks like Blizzard added a new value in the middle of the list, so the values being assigned in oUF to "powerName" and "powerTooltip" were the values Blizzard was naming "anchorTop" and "powerName", respectively.


ls-:
Well, I proposed two tiny altpower fixes https://github.com/haste/oUF/pull/252 few months ago :3

haste:
Merged ls-'s fixes.

Issue #257, opened 2015-09-30
Pet Frame not updating unit correctly


Sticklord:
Not sure if anyone keeps this updated anymore bute here we go:

When you're in a vehicle and the player jumps onto the petframe, and when you then exit again the player frame get stuck in the pet frame. (2 player frames). 

Deleting this line fixed it, but I'm sure it's there for a reason:
https://github.com/haste/oUF/blob/master/ouf.lua#L54

haste:
Is there a place I can test this?

Pondidum:
@haste I found the easiest place to test this is on the tillers farm with one of the weeds which you can yank up or get knocked around by.

Another option are the vehicles at the beginning of Ulduar.

p3lim:
I can't replicate this, tried with and without a pet frame, with and without a pet, at Tillers farm, Ulduar and Argent Tournament.

ls-:
@p3lim, I created a frost mage, went to Icecrown -> Argent Tournament, summoned my elemental, took a spear, got on horse, attacked **Melee Target** and dismounted, while in combat. My pet frame still displays player info.
http://imgur.com/a/0JqhN 
I moved pet frame to bottom right corner, so it's next to tooltip.

p3lim:
I see, got the same result.

Sticklord:
I'll just put some more info here. This is happening because UNIT_EXITED_VEHICLE gives player as arg1 on the petframe when the player leaves the vehicle, which makes sense because it was the player that left the vehicle. This unit (player) doesn't match either the realUnit or modUnit so the active unit on the pet frame does not get updated.

Rainrider:
@Sticklord is right about the reason it fails to update the pet frame on UNIT_EXITED_VEHICLE

I have some further questions though, just to be sure I'm not missing something.

In ouf.lua:
```lua
if(modUnit == "pet" and realUnit ~= "pet") then
	modUnit = "vehicle"
end
```
Is this really needed? E.g. `UnitHealth("pet")` and `UnitHealth("vehicle")` both return the same values when the player is in a vehicle.

```lua
if(not UnitExists(modUnit) or unit and unit ~= realUnit and unit ~= modUnit) then return end
```
What is the reason behind comparing the event unit with the frame units? Events are either registered by RegisterUnitEvent, so it will be the proper unit, or are unitless, so either there is no unit or the unit is not considered relevant.

p3lim:
[...]

PR #258, opened 2015-10-01
Fix petframe updating

Sticklord:
haste/oUF#257


Rainrider:
I think it would be better to just remove the check of the event unit against the frame units. It went fine in my limited testing. See last paragraph of https://github.com/oUF-wow/oUF/issues/257#issuecomment-246296377

Sticklord:
Okay cool, I see you got this.  I'll close this

Issue #260, opened 2015-11-13
Power/resource changes in Legion


p3lim:
This is just a compilation of all the information regarding power/resources for classes from the [Legion class previews](http://eu.battle.net/wow/en/blog/19941175), listing new/changed/removed powers/resources relevant to oUF.

## Additions

- Insanity
 - Shadow priests' new resource (replaces mana as primary resource)
 - Works like rage
 - Bar type element
- Lunar Power
 - Balance druids' new resource (replaces mana as primary resource)
 - Works like rage
 - Replaces Eclipse
 - Bar type element
- Arcane Charges
 - Arcane mages' "new" resource (class power)
 - It's basically the debuff turned into a proper class power
 - Pill type element
- Maelstrom
 - Elemental & enhancement shamans' new resource (replaces mana as primary resource)
 - Works like rage
 - Bar type element

## Changes

- Runes
 - Only one type now for all death knight specializations
 - Pill + bar type element (same as before)
- Chi
 - Only for windwalker monks
 - Pill type element
- Holy Power
 - Only for retribution paladins
 - Pill type element
- Soul Shards
 - Shared resource for all warlock specializations
 - Pill type element

## Removals

- Shadow Orbs
- Eclipse
- Holy Power (for protection and holy paladins)
- Chi (for brewmaster and mistweaver monks)
- Demonic Fury (although never implemented in oUF)
- Burning Embers (although never implemented in oUF)

## Notes

The following classes/specializations does not use mana as their primary resource any more.

- Shadow priests
- Balance druids
- Elemental & enhancement shamans
- Protection & retribution paladins

p3lim:
With these changes, some resources can finally be merged into single elements (pill and bar).

- Pill (classicons element)
 - Combo Points (should have been merged into this element already, since it's not per-target any more)
 - Chi (already implemented, but will need specialization checks)
 - Soul Shards (already implemented, but will need to remove specialization checks)
 - Shadow Orbs (needs to be removed)
 - Arcane Charges
- Bar (classpowers (?) element)
 - Insanity
 - Astral Power
 - Maelstrom
 - Druid Mana (should be merged into this element)

The eclipsebar element needs to be removed.  
The runebar element needs to be changed, and depending on what changes has been made to them, could potentially be merged into the pills (classicons) element.  
The totems element might need some changes, won't know until beta.

Not considered is the Demon Hunter, as I have no idea what resources they are using.

p3lim:
[...]

Issue #183, opened 2013-03-26
Updated ClassIcons for Destro Warlock


PProvost:
Hey haste,

I made some mods to classicons.lua so the same soul-shards icon system works for destruction lock burning embers.

You can see the edits here:
https://github.com/PProvost/ouf_quaiche/commit/624d52223309bee06397a756e48060c8da214b7d

If you'd prefer I can do a clone and pull request, but given the elegance of your lua code I suspect you'll have a neater way to do it than I did.

Regards,

PProvost

p3lim:
No longer needed in Legion, see #260.

haste:
On one side Legion won't be out until Q2/Q3 next year, so closing the issue is kinda stupid, but on the other hand... We're still not supporting it after so many years. :P

p3lim:
Was adding the note mostly to link the two issues together so it won't be forgotten once Legion came :)

PR #261, opened 2015-11-24
Legion patches

p3lim:
A continuation of #260 and eliminates the need of #183 come Legion.

This contains changes to the classicons and related elements, can't do much else without testing it in-game on each class.

All of the changes have fallback methods to the live game.

I will keep on pushing to this branch as I find more things that needs changes/fixes.


p3lim:
Suggestion: merging all the class/spec-specific tags into one (call it "classpower" maybe).

Would avoid many instances of having tag fonstrings being polled onupdate instead of by events simply because there are this many of them.

Gethe:
I merged this into my local version for testing on Alpha. When I logged in I got this error:
``` lua
1x ...terface\nibRealUI\Libs\oUF\elements\power.lua:118: attempt to index field 'Private' (a nil value)
```
The Private table is set to nil before any of the elements get loaded. What I've done for the moment is move the load of `finalize.lua` to after the last element, `elements\stagger.lua`.

Other than that, the changes seem to work fairly well, though I've only tested the classicons so far.

p3lim:
I'll fix that, the entire PR is drycoded since I don't have access to the alpha/beta yet.

Rainrider:
Just change
'local oUF = ns.oUF'
to
'local oUF = ns.oUF or oUF'?

p3lim:
Nah, I'll have to move the beta variable away from the private namespace, basically refactoring the PR.

Gethe:
I forgot one thing actually, the `UpdateTexture` function in classicons does not have a condition for Mages (Arcane Charges) or for Rogue/Druid (Combo Points)

Gethe:
After additional testing, I've isolated the following issues:

**classicons**
* As mentioned in my previous comment, `UpdateTexture` does not account for combo points or arcane charges.

* Word of Glory is a talent in Legion so it is no longer a reliable requirement; however, since only Retribution has Holy Power, just checking spec should suffice.

* SPELL_POWER_COMBO_POINTS is available on live realms and can get points from UnitPower(Max), so there is no need to have an `isBetaClient` check for combo points. This also means that the cpoints element can be deprecated *now* in favor of classicons

**runebar**
* There is a typo in the `UpdateRune` function where `runes` is indexed with `i` instead of `rid`

**tags**
* The spec check in `holypower` is using `==` when it should be `~=`.

**power**
* This isn't really an issue with oUF so much as it is with Blizzard. The indexed portion of `PowerBarColor` (and by extension `oUF.colors.power`) is not 100% accurate.
  - Combo Points should be at index 4
  - Chi is supposed to be at index 12
  - Arcane Charges are missing from index 16

  I have reported this to Blizz so if they decide to fix it, this portion will need changes too.


[This patch](https://github.com/haste/oUF/files/244258/patch.zip) contains the changes I've made. It has been tested on both Live and Alpha.

p3lim:
> Word of Glory is a talent in Legion so it is no longer a reliable requirement; however, since only Retribution has Holy Power, just checking spec should suffice.

The reason for checking Word of Glory is/was because it's the trigger spell (from a specific level) where Holy Power becomes relevant.

I'm going to work on this PR soon, preferably with access to the beta servers so I don't have to drycode everything.

Gethe:
> The reason for checking Word of Glory is/was because it's the trigger spell (from a specific level) where Holy Power becomes relevant.

Fair enough. If you intend to carry this forward, the first spell that uses Holy Power in Legion is [Templars Verdict](http://legion.wowhead.com/spell=85256/templars-verdict) at level 10.

p3lim:
Went over your patch and rebased it into the PR, thanks for testing!
If you find more spells like with Templar's Verdict, let me know.

Gethe:
I was actually looking at some of the early spells and found that Crusader Strike doesn't even give Holy Power until level 10, and is the only generator until 16. So my original solution is still valid, but it works either way.

p3lim:
Found a few more issues; tapping is deprecated (in the sense of how it used to work), and there seems to be an issue with portrait updates. Will test more when I get back home next week.

Gethe:
Another thing is the DruidMana element should also support dps specs for Priest and Shaman, who now have a different primary resource, but still have mana for heals and utility.

p3lim:
> Another thing is the DruidMana element should also support dps specs for Priest and Shaman, who now have a different primary resource, but still have mana for heals and utility.

Should already be covered by the element, although it could do with a new name (not in the scope of this PR? @haste comment on this)

Gethe:
> Should already be covered by the element, although it could do with a new name (not in the scope of this PR?

True, though it still has a hard return at the top of the file if you're not a druid, and the class color option assumes Druid.

p3lim:
@Gethe Pretty much rewrote DruidMana, used a power pairing table from the default UI: https://github.com/Gethe/wow-ui-source/blob/9396ea5331e1bd79fda2a5d32ea4668eb25e58ae/FrameXML/AlternatePowerBar.lua#L69-L79

ls-:
@p3lim, hi! I found couple of issues. First one is [here](https://github.com/p3lim-wow/oUF/blob/legion/elements/classicons.lua#L260), I dun think we should update textures, if icons aren't textures. There's also an issue w/ combo points and druids, only feral druids use them, but right now combo bar is enabled for all druid specs.

ls-:
Another one, stagger bar doesn't work. Can be fixed by replacing ```STANCE_OF_THE_STURY_OX_ID ~= GetShapeshiftFormID()``` with ```SPEC_MONK_BREWMASTER ~= GetSpecialization()``` [here](https://github.com/p3lim-wow/oUF/blob/master/elements/stagger.lua#L119).

p3lim:
@ls- Thanks, will look into both tomorrow.

ls-:
BTW, there's a mana cost prediction right now, why don't we add one as an element akin to heal prediction?

p3lim:
> There's also an issue w/ combo points and druids, only feral druids use them, but right now combo bar is enabled for all druid specs.

The 3rd tier of talents for druids allow them to use parts of other specializations, so choosing Feral Affinity there gives any druid spec cat abilities.  
Nevertheless, the element would not show combo points unless the player has gotten some, so there shouldn't be any issue.  
If the layout _really_ wants to mimic the default unitframes they can handle that themselves, but (in my opinion) it's not something that oUF should handle.

Example implementation:
```Lua
local function UpdateDruidForm(self, event)
	local element = self.ClassIcons

	local shouldShow = GetShapeshiftFormID() == CAT_FORM
	local cur = UnitPower('player', SPELL_POWER_COMBO_POINTS)
	for i = 1, element.__max do
		if(shouldShow and i <= cur) then
			element[i]:Show()
		else
			element[i]:Hide()
		end
	end
end

-- in style
if(select(2, UnitClass('player')) == 'DRUID') then
	self:RegisterEvent('UPDATE_SHAPESHIFT_FORM', UpdateDruidForm)
end
```

p3lim:
> BTW, there's a mana cost prediction right now, why don't we add one as an element akin to heal prediction?

Will have to look at the default code for that, maybe something that could be considered.

p3lim:
Also, since combo points was introduced in classicons, the cpoints element could (and IMO should) be removed.

ls-:
> Will have to look at the default code for that, maybe something that could be considered.

Function code is [here](https://github.com/Gethe/wow-ui-source/blob/beta/FrameXML/UnitFrame.lua#L373-L394), and [here](https://github.com/Gethe/wow-ui-source/blob/beta/FrameXML/UnitFrame.lua#L220-L222)'s how it's called. IMHO, we should add it. I may right pcp element tomorrow.

Rainrider:
Combo points are not tied to druids and rogues only. There is [Aces High!](http://www.wowhead.com/quest=13413/aces-high) and I think the Malygos encounter and one more in The Nexus where you fight in a vehicle. This is also bugged on Beta right now, because UnitPower("vehicle", SPELL_POWER_COMBO_POINTS) always returns 0 despite of UNIT_POWER_FREQUENT firing for the vehicle unit. GetComboPoints("vehicle") returns correct results. The default ui fails to handle that for other reasons too.

ls-:
Ah, damn, always forget about vehicle combo points T_T Haven't seen them in a while...

ls-:
pic

So today I wrote something like this, I also made it work w/ alt power source a.k.a. DruidMana bar.

There are two ways to implement it:
A) Akin to heal prediction element, and that's how I did it, a table containing two status bars, i.e. mainBar and altBar, mainBar is used as power cost prediction bar for main power source bar, and altBar is used for alternative one, if any.
B) As a single status bar, which can be re-anchored in layout. TBH, I'm not sure that scenario w/ two visible pcp bars is even possible.

Tell me which way to do it, and I'll create a proper pull request. Or I'll make it into oUF plugin :grin:

p3lim:
@ls- Does it alter the original power element or adds to it?

Rainrider:
It adds a new texture on top of it. Pretty much like the (heal) absorb bar

ls-:
@p3lim, what @Rainrider said. A new bar on top of it. For now I implemented it as a set of two status bars, cuz I was a bit lazy while drycoding, hehe, but I was going to rewrite them as textures anyway.

Sooo.. What should I do? 
A) One texture for two bars, that needs to be re-anchored in element's post-update by layout;
B) Two textures, that need to be anchored only once at their creation;
C) Make it completely blizzlike, one texture for main power bar only.

PCP doesn't change any other oUF element and is absolutely independent, can be anchored to anything, even centre of the screen :D 

Current update code:
```lua
local function Update(self, event, unit)
    if(self.unit ~= unit) then return end

    local pcp = self.PowerCostPrediction

    if(pcp.PreUpdate) then
        pcp:PreUpdate(unit)
    end

    local _, _, _, _, startTime, endTime, _, _, _, spellID = UnitCastingInfo(unit)
    local mainPowerType = UnitPowerType(unit)
    local hasAltPowerBar = ALT_MANA_BAR_PAIR_DISPLAY_INFO[playerClass] and ALT_MANA_BAR_PAIR_DISPLAY_INFO[playerClass][mainPowerType]
    local isStarting = event == "UNIT_SPELLCAST_START"
    local mainCost, altCost = 0, 0

    if(isStarting or startTime ~= endTime) then
        local costTable = GetSpellPowerCost(spellID)

        for _, costInfo in pairs(costTable) do
            --[[costInfo content:
                -- name: string (powerToken)
                -- type: number (powerType)
                -- cost: number
                -- costPercent: number
                -- costPerSec: number
                -- minCost: number
                -- hasRequiredAura: boolean
                -- requiredAuraID: number
            ]]
            if(costInfo.type == mainPowerType) then
                mainCost = costInfo.cost

                break
            elseif(costInfo.type == ADDITIONAL_POWER_BAR_INDEX) then
                altCost = costInfo.cost

                break
            end
        end
    end

    if(pcp.mainBar) then
        pcp.mainBar:SetMinMaxValues(0, UnitPowerMax(unit, mainPowerType))
        pcp.mainBar:SetValue(mainCost)
        pcp.mainBar:Show()
    end

    if(pcp.altBar and hasAltPowerBar) then
        pcp.altBar:SetMinMaxValues(0, UnitPowerMax(unit, ADDITIONAL_POWER_BAR_INDEX))
        pcp.altBar:SetValue(altCost)
        pcp.altBar:Show()
    end

    if(pcp.PostUpdate) then
        return pcp:PostUpdate(unit, mainCost, altCost, hasAltPowerBar)
    end
end
```

p3lim:
> This is also bugged on Beta right now, because UnitPower("vehicle", SPELL_POWER_COMBO_POINTS) always returns 0 despite of UNIT_POWER_FREQUENT firing for the vehicle unit.

@Rainrider It's like that on live too, oddly enough `UnitPowerMax('vehicle', SPELL_POWER_COMBO_POINTS)` works fine.

ls-:
@p3lim, could you please add powerType arg to classicons' PostUpdate? O_o That'd be really nice.

ls-:
Oh, about runebar, [this block](https://github.com/p3lim-wow/oUF/blob/legion/elements/runebar.lua#L99-L103) should be changed, cuz it prevents PostUpdate callback from being called.
```Lua
if(UnitHasVehicleUI'player') then
	rune:Hide()
else
	rune:Show()

	local start, duration, runeReady = GetRuneCooldown(rid)
	if(runeReady) then
		rune:SetMinMaxValues(0, 1)
		rune:SetValue(1)
		rune:SetScript("OnUpdate", nil)
	else
		rune.duration = GetTime() - start
		rune.max = duration
		rune:SetMinMaxValues(1, duration)
		rune:SetScript("OnUpdate", OnUpdate)
	end
end
```
Something like this might be a better solution.

Rainrider:
Is there a particular reason behind all the beta build checks and the maintained compatibility with the live master branch?

It becomes increasingly difficult to track legion changes und this issue is already hard enough to get hold off because of all the different topics that arise.

Wouldn't it be easier if we either did PRs against p3lim's legion branch or if we had an official beta branch like it used to be?

Rainrider:
@ls- I think the whole UpdateRune function is redundant and should be moved to Update. Also calling UnitHasVehicleUI 6 times instead of 1 is a waste.

p3lim:
@Rainrider The reasoning behind the checks is that there is one version that is compatible with both clients, rather than separating the project into two versions we'd (or rather haste) has to track.

Having a specific branch we could base our PRs against would be nice, but it'd have to be on the original repository, not a fork like mine, it'd be hard for everyone to track that.
But that is all up to haste, he has yet to input on anything in this PR.

p3lim:
> @ls- I think the whole UpdateRune function is redundant and should be moved to Update. Also calling UnitHasVehicleUI 6 times instead of 1 is a waste.

I agree, the runes have been simplified a lot the past few expansions, a cleanup would do well.

Rainrider:
> The reasoning behind the checks is that there is one version that is compatible with both clients, rather than separating the project into two versions we'd (or rather haste) has to track.
> 
> Having a specific branch we could base our PRs against would be nice, but it'd have to be on the original repository, not a fork like mine, it'd be hard for everyone to track that.
> But that is all up to haste, he has yet to input on anything in this PR.

I think maintaining compatibility between live and beta is just more work. Add compat first and remove it later. Having a beta branch makes it all much lighter - less conditional branching, easier maintaining, no constraints on renaming elements (DruidMana, ClassIcons). When Legion gets released just merge the beta branch and all is done.

Also I'm not sure if I am doing it right and welcome comments on that, but currently I have like 5 different remotes on my local oUF repo, a legion branch that tracks p3lim's legion branch and another legion branch where I merge p3lim's and ls-'s changes (which I track manually through mail notifications when the relevant PRs are updated). If someone does a rebase, I throw everything away and start over. This is an overkill.

p3lim:
> Also I'm not sure if I am doing it right and welcome comments on that, but currently I have like 5 different remotes on my local oUF repo, a legion branch that tracks p3lim's legion branch and another legion branch where I merge p3lim's and ls-'s changes (which I track manually through mail notifications when the relevant PRs are updated). If someone does a rebase, I throw everything away and start over. This is an overkill.

This all assumes that PRs get constantly merged, if they don't you still have to go through the same process.

ls-:
Heh, so I'm not the only one who thinks that runebar need an overhaul. That's good.

Yeah, managing many branches may be pain in the ass, I also feel sorry for anyone who tracks my reps T_T Today I pruned couple of branches and rebased many things.

p3lim:
> I think the whole UpdateRune function is redundant and should be moved to Update. Also calling UnitHasVehicleUI 6 times instead of 1 is a waste.

@Rainrider Actually, because of how _RUNE_POWER_UPDATE_ works (per-rune updates) and how we need to iterate over all 6 when we force-update (PEW/RefreshUnit/ForceUpdate), and if we want to provide `PreUpdate`, calling `UnitHasVehicleUI` 6 times at once is the better solution, and it's a rare occurance anyways.

p3lim:
This PR has gotten to the point that it's hard to work out what has changed or not, and having 40+ comments does not help haste at all.

So, upon request I've split it up into multiple PRs, each for their own changes/element.

Power colors: #267 
Tapping: #264 
Tags: #265 
ClassIcons: #266 
DruidMana: #268 
Runes: #269 
Eclipse: #270 
Stagger: #271 
CPoints: #272

PR #262, opened 2016-05-26
New power cost prediction element.

ls-:
Here's my take on PCP.

I haven't added element description yet, cuz current implementation isn't set in stone, I'll add it when everything else is done.

It works quite similar to heal prediction element.
```lua
local mainBar = CreateFrame('StatusBar', nil, self.Power)
mainBar:SetReverseFill(true)
mainBar:SetPoint('TOP')
mainBar:SetPoint('BOTTOM')
mainBar:SetPoint('RIGHT', self.Power:GetStatusBarTexture(), 'RIGHT')
mainBar:SetWidth(200)

local altBar = CreateFrame('StatusBar', nil, self.DruidMana)
altBar:SetReverseFill(true)
altBar:SetPoint('TOP')
altBar:SetPoint('BOTTOM')
altBar:SetPoint('RIGHT', self.DruidMana:GetStatusBarTexture(), 'RIGHT')
altBar:SetWidth(200)

self.PowerCostPrediction = {
    mainBar = mainBar,
    altBar = altBar
}
```

As I said [here](https://github.com/haste/oUF/pull/261#issuecomment-221647406), there're 3 ways to implement it:

1. One texture/statusbar for two bars, that needs to be re-anchored in element's post-update by layout;
2. Two textures/statusbars, that need to be anchored only once at their creation. Current version;
3. Make it completely blizzlike, one texture for main power bar only.

1 and 2 are almost identical, 2 actually covers 1, both bars are optional, so you can create one bar and override internal update process. However 2 is a bit of overkill, I made it this way, cuz theoretically there's a chance, that both bars will be displayed.

Right now both sub-widgets are statusbars, cuz they're easier to work with, but can be re-implemented as textures, if you, guys, think that will be a better way of doing things.

Here are examples of PCP widget w/ oUF_P3lim (tiny grey segment on power bar)
image
and w/ my UI (white vertical bars)
image

### DIFF
--- a/elements/powerprediction.lua
+++ b/elements/powerprediction.lua
@@ -47,8 +47,6 @@
                   to its internal function again.
 ]]
 
-if(select(4, GetBuildInfo()) < 70000) then return end
-
 local _, ns = ...
 local oUF = ns.oUF
 
--- a/elements/powerprediction.lua
+++ b/elements/powerprediction.lua
@@ -1,0 +1,161 @@
+--[[ Element: Power Prediction Bar
+ Handles updating and visibility of the power prediction status bars.
+
+ Widget
+
+ PowerPrediction - A table containing `mainBar` and `altBar`.
+
+ Sub-Widgets
+
+ mainBar - A StatusBar used to represent power cost of spells, that consume
+           your main power, e.g. mana for mages;
+ altBar  - A StatusBar used to represent power cost of spells, that consume
+           your additional power, e.g. mana for balance druids.
+
+ Notes
+
+ The default StatusBar texture will be applied if the UI widget doesn't have a
+ status bar texture.
+
+ Examples
+
+   -- Position and size
+   local mainBar = CreateFrame('StatusBar', nil, self.Power)
+   mainBar:SetReverseFill(true)
+   mainBar:SetPoint('TOP')
+   mainBar:SetPoint('BOTTOM')
+   mainBar:SetPoint('RIGHT', self.Power:GetStatusBarTexture(), 'RIGHT')
+   mainBar:SetWidth(200)
+
+   local altBar = CreateFrame('StatusBar', nil, self.DruidMana)
+   altBar:SetReverseFill(true)
+   altBar:SetPoint('TOP')
+   altBar:SetPoint('BOTTOM')
+   altBar:SetPoint('RIGHT', self.DruidMana:GetStatusBarTexture(), 'RIGHT')
+   altBar:SetWidth(200)
+
+   -- Register with oUF
+   self.PowerPrediction = {
+      mainBar = mainBar,
+      altBar = altBar
+   }
+
+ Hooks
+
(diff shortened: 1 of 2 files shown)


p3lim:
Looks nice :+1:

haste:
Couldn't it be named PowerPredition, slightly in the style of HealPrediction?

It would also be nice if you could rebase your commits and prepend the name of the element. Like most other commits are styled. That way I can just merge it through github :)

ls-:
@haste, done :+1:

elements/powerprediction.lua
```diff
@@ -0,0 +1,161 @@
...
+
+   -- Position and size
+   local mainBar = CreateFrame('StatusBar', nil, self.Power)
+   mainBar:SetReverseFill(true)
+   mainBar:SetPoint('TOP')
+   mainBar:SetPoint('BOTTOM')
+   mainBar:SetPoint('RIGHT', self.Power:GetStatusBarTexture(), 'RIGHT')
+   mainBar:SetWidth(200)
+
+   local altBar = CreateFrame('StatusBar', nil, self.DruidMana)
```
p3lim: Typo

elements/powerprediction.lua
```diff
@@ -0,0 +1,161 @@
...
+   altBar:SetReverseFill(true)
+   altBar:SetPoint('TOP')
+   altBar:SetPoint('BOTTOM')
+   altBar:SetPoint('RIGHT', self.DruidMana:GetStatusBarTexture(), 'RIGHT')
```
p3lim: Same here.

p3lim:
The element doesn't seem to work correctly.
Using the example code (only added color to make it stand out for the screenshots), the PCP bar fills up most of the power bar, and when I start to spend power it expands beyond the power bar to the left.

image
image

elements/powerprediction.lua
Rainrider: It seems right how Val did it. E.g. shamans have maelstrom as their primary resource, which is displayed on the standard powerbar, and mana as a secondary resource for heals, which oUF would display as the DruidMana element. Maelstrom spells would drain the main powerbar (self.Power) and heals would drain the secondary one (self.DruidMana).

p3lim:
Testing it out on a target dummy with a hunter, saw some weird behavior compared to the playernameplate:
http://i.imgur.com/Omt3pzu.gifv

Scaled the playerframe up a bit, the red bit is this element.

ls-:
Hmm... That's weird, could you show me how you created and anchored it? So I can reproduce it.

I've been playing w/ it for quite a long time, never experienced anything like that O_o

p3lim:
I've got pretty much the exact same code as the one you provided there.
Oddly enough, today it worked just fine, I can't reproduce the issue.

p3lim:
Please remove the compatibility checks, Legion is upon us already.

PR #263, opened 2016-05-27
Prestige icon.

ls-:
And another one.

Works this way:
```Lua
   -- Position and size
   local PvP = self:CreateTexture(nil, 'ARTWORK', nil, 1)
   PvP:SetSize(30, 30)
   PvP:SetPoint('RIGHT', self, 'LEFT')

   local Prestige = self:CreateTexture(nil, 'ARTWORK')
   Prestige:SetSize(50, 52)
   Prestige:SetPoint('CENTER', PvP, 'CENTER')

   -- Register it with oUF
   self.PvP = PvP
   self.PvP.Prestige = Prestige
```

Looks this way:
image

Merged pvp and prestige elements. If `Prestige` sub-widget isn't defined or prestige level is 0, PvP icon will be shown.

I also changed `PostUpdate`.
Live:
```Lua
pvp:PostUpdate(status)
```
Beta:
```Lua
pvp:PostUpdate(unit, status, hasPrestige, level)
```


p3lim:
I'd say merge them, showing the prestige badge instead of the normal pvp icon if the unit has a prestige level.
It should also be optional, some may not want it.

ls-:
@p3lim, prestige emblem consists of two elements: portrait - background image (hexagon) and badge - icon itself. I decided to preserve blizz names for these elements.

p3lim:
What about this:
If the layout only supplies the icon, it will only show the old PvP icons (and perhaps mercenary, don't know how that works), and if the layout also supplies a "Badge" as a sub-element (opposite of how you suggested) it should show the whole prestige thing too (making the previous PvP element the portrait).

Rainrider:
> I'm taking a few notes from your PCP element in my next update to the Experience plugin because of this

@p3lim I already have something like that based on your Experience plugin:
[The element](https://github.com/Rainrider/oUF_Rain/blob/master/modules/rain_experience.lua) and the [layout code](https://github.com/Rainrider/oUF_Rain/blob/master/rain_functions.lua#L854-L896). I see I could actually move the positioning into the element itself.

ls-:
> if the layout also supplies a "Badge" as a sub-element (opposite of how you suggested) it should show the whole prestige thing too (making the previous PvP element the portrait).

I was thinking about it too, but It may be a bit problematic to do though. Portraits (hexagons) are quite BIG in size, they are 50px p 52px and aren't even squares :hankey:.

> The only reason I suggest this is so that we don't have to force the layouts to mess around with sublayers or any other required methods of creating the element

However, we may have to do so. If you don't specify them, they'll all be drawn in same 0 sublayer, and then one of them will randomly appear to be on top of another one. Setting sublayers manually is pretty much mandatory.
image
It took me 1 /reload to get portrait to be drawn on top of badge, although it's created first. This pic also shows how big portraits are.

ls-:
Here we go.
```Lua
   -- Position and size
   local PvP = self:CreateTexture(nil, 'ARTWORK', nil, 1)
   PvP:SetSize(30, 30)
   PvP:SetPoint('RIGHT', self, 'LEFT')

   local Prestige = self:CreateTexture(nil, 'ARTWORK')
   Prestige:SetSize(50, 52)
   Prestige:SetPoint('CENTER', PvP, 'CENTER')

   -- Register it with oUF
   self.PvP = PvP
   self.PvP.Prestige = Prestige
```
I also updated element description, added note about sub-layers, cuz they have to be defined, there's no other way.

Now it looks this way in game.
image

p3lim:
:+1:

p3lim:
Please remove the compatibility checks, Legion is upon us already.

PR #265, opened 2016-05-31
Legion tags

p3lim:
This PR contains additions/deprecations for the tags element for Legion.

### DIFF
--- a/elements/tags.lua
+++ b/elements/tags.lua
@@ -281,11 +281,9 @@ local tagStrings = {
 	end]],
 
 	['soulshards'] = [[function()
-		if(IsPlayerSpell(WARLOCK_SOULBURN)) then
-			local num = UnitPower('player', SPELL_POWER_SOUL_SHARDS)
-			if(num > 0) then
-				return num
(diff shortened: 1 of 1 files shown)


ls-:
While I was merging all PRs today, found one issue w/ this commit, more like a typo. It doesn't break anything. I dunno why I'm even writing this comment 😆 

Soulshards are mentioned few times, but entry [at line 423](https://github.com/p3lim-wow/oUF/blob/legion-tags/elements/tags.lua#L423) isn't needed.

p3lim:
@ls- I'm gonna let that one slide until the post-patch cleanup.

p3lim:
Removed the compatibility code since 7.0 is already out.

ls-:
You forgot to removed ['pereclipse'] tag from [Here](https://github.com/p3lim-wow/oUF/blob/legion-tags/elements/tags.lua#L399) and [here](https://github.com/p3lim-wow/oUF/blob/legion-tags/elements/tags.lua#L266-L273).

PR #266, opened 2016-05-31
Legion classicons

p3lim:
This PR contains additions/deprecations for the classicons element for Legion.
It depends on #267 to work correctly for Legion.
It also eliminates the need for #223 and #240.

### DIFF
--- a/elements/classicons.lua
+++ b/elements/classicons.lua
@@ -214,7 +214,12 @@ do
 	elseif(PlayerClass == 'PALADIN') then
 		ClassPowerID = SPELL_POWER_HOLY_POWER
 		ClassPowerType = "HOLY_POWER"
-		RequireSpell = 85673 -- Word of Glory
+
+		if(isBetaClient) then
+			RequireSpec = SPEC_PALADIN_RETRIBUTION
+		else
+			RequireSpell = 85673 -- Word of Glory
+		end
 	elseif(PlayerClass == 'PRIEST') then
 		ClassPowerID = SPELL_POWER_SHADOW_ORBS
 		ClassPowerType = "SHADOW_ORBS"
--- a/elements/classicons.lua
+++ b/elements/classicons.lua
@@ -12,7 +12,6 @@
 
  Monk    - Chi Orbs
  Paladin - Holy Power
- Priest  - Shadow Orbs
  Warlock - Soul Shards
 
  Examples
@@ -220,7 +219,7 @@ do
 		else
 			RequireSpell = 85673 -- Word of Glory
 		end
-	elseif(PlayerClass == 'PRIEST') then
+	elseif(PlayerClass == 'PRIEST' and not isBetaClient) then
 		ClassPowerID = SPELL_POWER_SHADOW_ORBS
 		ClassPowerType = "SHADOW_ORBS"
 		RequireSpec = SPEC_PRIEST_SHADOW
--- a/elements/classicons.lua
+++ b/elements/classicons.lua
@@ -50,6 +50,8 @@
 local parent, ns = ...
 local oUF = ns.oUF
 
+local isBetaClient = select(4, GetBuildInfo()) >= 70000
+
 local _, PlayerClass = UnitClass'player'
 
 -- Holds the class specific stuff.
@@ -205,6 +207,10 @@ do
 	if(PlayerClass == 'MONK') then
 		ClassPowerID = SPELL_POWER_CHI
 		ClassPowerType = "CHI"
+
+		if(isBetaClient) then
+			RequireSpec = SPEC_MONK_WINDWALKER
+		end
 	elseif(PlayerClass == 'PALADIN') then
 		ClassPowerID = SPELL_POWER_HOLY_POWER
 		ClassPowerType = "HOLY_POWER"
--- a/elements/classicons.lua
+++ b/elements/classicons.lua
@@ -227,8 +227,11 @@ do
 	elseif(PlayerClass == 'WARLOCK') then
 		ClassPowerID = SPELL_POWER_SOUL_SHARDS
 		ClassPowerType = "SOUL_SHARDS"
-		RequireSpec = SPEC_WARLOCK_AFFLICTION
-		RequireSpell = WARLOCK_SOULBURN
+
+		if(not isBetaClient) then
+			RequireSpec = SPEC_WARLOCK_AFFLICTION
+			RequireSpell = WARLOCK_SOULBURN
+		end
 	end
 end
 
--- a/elements/classicons.lua
+++ b/elements/classicons.lua
@@ -62,27 +62,14 @@ local ClassPowerEnable, ClassPowerDisable
 local RequireSpec, RequireSpell
 
 local UpdateTexture = function(element)
-	local red, green, blue, desaturated
-	if(PlayerClass == 'MONK') then
-		red, green, blue = 0, 1, .59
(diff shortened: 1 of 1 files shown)


ls-:
As for druid's combo points. We have requireSpec, requireSpell, why dun we add requirePower to handle [this case](https://github.com/Gethe/wow-ui-source/blob/beta/FrameXML/ComboFramePlayer.lua#L51)? :laughing:

ls-:
Hey, I made a small patch for this PR https://github.com/ls-/oUF/commit/cd20ebc03949a2feaa0d8ef3c6d2c4fc0c72631f. It fixes combo points for dudus, e.g. combo points are only shown for cat form, so I went blizz way and added new `RequirePower` check. If you dun like it, no problem, I'll keep it to myself, hehe.

On a different note, `UNIT_DISPLAYPOWER` event has no arguments besides unit, so there's no need to register it for `Path`, cuz it'll never get into `Update` function anyway. IMHO, it should be used for `VisibilityPath`.

ls-:
There's one more issue w/ class icons. Can't apply fix it to this PR, cuz it's not based on latest master commit 😕, but this issue was introduced by your PR.

Please, change [this line](https://github.com/haste/oUF/blob/master/elements/classicons.lua#L172) to
```Lua
Path(self, event, unit, unit == 'vehicle' and 'COMBO_POINTS' or ClassPowerType)
```
after you rebase this PR.

Rainrider:
You could do the pull request against p3lim's branch. Of course you should
apply your changes onto his branch as well.

>

ls-:
@Rainrider, I can, sure, but at first I wanna know if needs these changes in his PR or not, He may say that it should be handled by layout 😕 

And second fix can't be applied to his branch.

p3lim:
Any element should mimic the blizzard behavior by standard, and even if I personally prefer to see the combo points outside cat form, it should be merged.

Feel free to send the PR my way.

p3lim:
Rebased the tip of the branch to master, no idea why it wasn't already.

p3lim:
Found one more issue while testing on Malygos, `GetComboPoints` is not ready on `PLAYER_ENTERING_WORLD` (reloading UI) and will report `0` regardless.

Not sure if we want to properly support such a case with custom logic, as it's only appearing in these rare situations when we're testing, and GetComboPoints is/should be deprecated anyhow, just up to blizzard to fix UnitPower

ls-:
TBH, there're quite many things that aren't ready on PLAYER_ENTERING_WORLD in Legion. For example, there's an issues w/ shapeshifting. If you log in or reload UI in some sort of form, many functions will return values for humanoid form.

Even blizz UI suffers from this issue.
image
Here I logged in as a cat, yet I have Lunar energy and additional mana bars. I dunno what they did to loading process in Legion.

Blazeflack:
When you added the "RequirePower" check, I think it broke it a bit for druids going in and out of cat form. If you log in while in cat form, the combo points will show. But if you then go out of cat form, the visibility update will disable the element and unregister "UNIT_DISPLAYPOWER" event, so it will never be enabled again.

ls-:
@Blazeflack , when `UNIT_DISPLAYPOWER` is disabled, `SPELLS_CHANGED` is used for visibility checks, it fires whenever players shapeshifts.

Blazeflack:
I'm not seeing the "SPELLS_CHANGED" event fire at all:
http://i.imgur.com/00iZmcY.jpg

ls-:
Have no issues "SPELLS_CHANGED" :(
image
Here I logged in as a cat, switched to travel form and switched back to cat.

Blazeflack:
Did you see the edits I made to my post?

ls-:
Hm, yeah. But this `SPELL_CHANGED` issue is really weird. It occurs in you case, but it doesn't happen in mine, although we hopefully run same version of oUF. That's why I posted everything I posted 😐 

And thank you, although I need to tweak texture a bit T_T BTW, I'm about to steal some actionbar code from ElvUI 😁 Ofc I'll mention from it's stolen.

Blazeflack:
Yeah I compared the classicons files just to make sure, and it matches 100% to the file in p3lims branch.

I think adding a spell check for Shred is a good way to alleviate most of the issues surrounding combopoints in various druid forms.

Blazeflack:
I made the following changes, and it seems to work a lot better:
`From cda7671ab401b056185aa7af1a40f4e59c16aa85 Mon Sep 17 00:00:00 2001
From: Blazeflack <blazeflack@hotmail.com>
Date: Mon, 18 Jul 2016 11:36:11 +0200
Subject: [PATCH] Added Shred as RequireSpell for Druids in the ClassIcons
 element

---
 elements/classicons.lua | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/elements/classicons.lua b/elements/classicons.lua
index c0ac3ea..cb01a37 100644
--- a/elements/classicons.lua
+++ b/elements/classicons.lua
@@ -161,6 +161,8 @@ local function Visibility(self, event, unit)
 				else
 					self:RegisterEvent('SPELLS_CHANGED', Visibility, true)
 				end
+			elseif (RequireSpell) then
+				self:RegisterEvent('SPELLS_CHANGED', Visibility, true)
 			end
 		end
 	end
@@ -244,6 +246,7 @@ do
 
 		if(isBetaClient and PlayerClass == 'DRUID') then
 			RequirePower = SPELL_POWER_ENERGY
+			RequireSpell = 5221 -- Shred
 		end
 	elseif(PlayerClass == 'MAGE' and isBetaClient) then
 		ClassPowerID = SPELL_POWER_ARCANE_CHARGES
-- 
2.6.4.windows.1`

Much more consistency, the combopoints show up when I expect them to and hide when they should.

Blazeflack:
There is still the issue you mentioned here: https://github.com/haste/oUF/pull/266#issuecomment-233147444

Logging in or reloading while in a form returns values as if you were in human form. Not sure if there is any way we can work around that.

ls-:
That issue is unfixable 😞 Something is broken on blizz side, they even tried to solve by adding delayed update to their combo widget via `C_Timer`, but it didn't help, cuz functions start returning proper values **only after** player shapeshifts.

ls-:
As for your fix, `elseif (RequireSpell) then`, simple `else` is enough, cuz it'll go that way only if there's a required spell, but character didn't have it yet, and `IsPlayerSpell(RequireSpell)` returned false.

Blazeflack:
Oh you're right. Sometimes I fail at logic (or logic fails me) :(

p3lim:
I'm replacing the power logic for Shred check, there is no point allowing CP to be shown unless one can gain it.

ls-:
@p3lim, well, feral druids in humanoid form has Shred, but don't use focus, so they don't have combo bar and can't use said abilities, both checks are needed.

p3lim:
@ls- I would say that if you end up being in a situation where you're in humanoid form and still have combo points, it'd be in a situation where you'd want to use them soon anyways, and knowing exactly how many you have could be a good thing.

p3lim:
I'm happy to bring up that subject again at a later date, but right now I'm focusing on finishing the PRs so that @haste can (finally) review them, we're way too close to (and in some regions already past) the patch.

p3lim:
That'll be the last rebase of this PR for now.

ls-:
I still think, that @haste should add you @p3lim as a collaborators for oUF.

PR #268, opened 2016-05-31
Legion druidmana

p3lim:
This PR contains changes for the druidmana element for Legion, as more classes supports it.

These commits are compatible with both live and beta clients

### DIFF
--- a/elements/druidmana.lua
+++ b/elements/druidmana.lua
@@ -18,12 +18,14 @@
 
  Options
 
- .colorClass  - Use `self.colors.class[class]` to color the bar. This will
-                always use DRUID as class.
- .colorSmooth - Use `self.colors.smooth` to color the bar with a smooth
-                gradient based on the players current mana percentage.
- .colorPower  - Use `self.colors.power[token]` to color the bar. This will
-                always use MANA as token.
+ .colorClass   - Use `self.colors.class[class]` to color the bar. This will
+                 always use DRUID as class.
+ .colorSmooth  - Use `self.colors.smooth` to color the bar with a smooth
+                 gradient based on the players current mana percentage.
+ .colorPower   - Use `self.colors.power[token]` to color the bar. This will
+                 always use MANA as token.
+ .displayPairs - Overridable table of pairs used to match class and power to
+                 display or hide the element.
 
  Sub-Widget Options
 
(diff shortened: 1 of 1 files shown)


p3lim:
Considering the changes in legion, and to support any future additions to this feature (other powers perhaps?), I'd like to suggest we rename the element to either "AlternativePower" or "AdditionalPower" (I prefer the latter, and it's what's being used in the constants).

ls-:
I agree! :+1: DruidMana doesn't make sense anymore.

Blazeflack:
The following check is broken:
`if(UnitPowerType('player') == ADDITIONAL_POWER_BAR_NAME) then`

You want the 2nd return of UnitPowerType, so the following will fix it:
`if(select(2, UnitPowerType('player')) == ADDITIONAL_POWER_BAR_NAME) then`

p3lim:
@Blazeflack Or just match with `ADDITIONAL_POWER_BAR_INDEX`.

Blazeflack:
Oh yeah.

ls-:
There're few issues w/ new version of druidmana.
1) This [check](https://github.com/p3lim-wow/oUF/blob/legion-druidmana/elements/druidmana.lua#L76) doesn't make sense, it'll hide the bar only if player's power is MANA. I believe it should be something like this `not ALT_MANA_BAR_PAIR_DISPLAY_INFO[playerClass][UnitPowerType('player')]`;
2) Bailing out at [this point](https://github.com/p3lim-wow/oUF/blob/legion-druidmana/elements/druidmana.lua#L77) is bad, preventing PostUpdate from running is wrong 😠 💢 .

p3lim:
> I believe it should be something like this not `ALT_MANA_BAR_PAIR_DISPLAY_INFO[playerClass][UnitPowerType('player')]`.

https://github.com/p3lim-wow/oUF/blob/legion-druidmana/elements/druidmana.lua#L139-L163

The hiding in the Update function is probably something left over.

> Bailing out at this point is bad, preventing PostUpdate from running is wrong

Could just hook `OnHide` if you *really* need to know.

ls-:
Well, if it's leftover, then then everything is fine, cuz its deletion solves both issues :D

p3lim:
Everything should be handled by Visibility, if not it's a bug. Feel free to test, I'm busy today.

ls-:
Yeah, if [this block](https://github.com/p3lim-wow/oUF/blob/legion-druidmana/elements/druidmana.lua#L75-L80) is deleted, both issues are solved. I thought it's intended, thus I was confused 😁 

Have a nice day 👍

Blazeflack:
The "return additionalpower:Hide()" call in Update is called when you go into Flight Form as a Balance Druid. There may be other cases, but at least this one case is confirmed.

p3lim:
> The "return additionalpower:Hide()" call in Update is called when you go into Flight Form as a Balance Druid. There may be other cases, but at least this one case is confirmed.

I can't verify this on the beta.

PR #269, opened 2016-05-31
Legion runebar

p3lim:
This PR contains changes to the runebar element for Legion.
It also eliminates the need for #112.

### DIFF
--- a/elements/runebar.lua
+++ b/elements/runebar.lua
@@ -143,21 +143,11 @@ local Enable = function(self, unit)
 		self:RegisterEvent("RUNE_POWER_UPDATE", UpdateRune, true)
 		self:RegisterEvent("RUNE_TYPE_UPDATE", UpdateType, true)
 
-		-- oUF leaves the vehicle events registered on the player frame, so
-		-- buffs and such are correctly updated when entering/exiting vehicles.
-		--
-		-- This however makes the code also show/hide the RuneFrame.
-		RuneFrame.Show = RuneFrame.Hide
-		RuneFrame:Hide()
-
 		return true
 	end
 end
 
 local Disable = function(self)
-	RuneFrame.Show = nil
-	RuneFrame:Show()
-
 	self:UnregisterEvent("RUNE_POWER_UPDATE", UpdateRune)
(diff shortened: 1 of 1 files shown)


ls-:
Could you please address the issue I described [here](https://github.com/haste/oUF/pull/261#issuecomment-222419876)?

p3lim:
Sure, but it'll wait until tomorrow.

ls-:
No problem there. I decided to remind about it, cuz I wasn't sure whether or not you saw that comment. As you said, that PR had become too big, so that's quite possible :sweat_smile:

Blazeflack:
I think you may have unintentionally made some find replace errors in the "Add Override support and rename the PostUpdate hook" commit.

All references to the "UpdateRune" function was replaced with "Update", most likely during your replacement of "PostUpdateRune" with "PostUpdate".

Blazeflack:
I forgot to mention that this probably breaks the element, as there is now two functions named "Update", so one replaces the other.

p3lim:
@Blazeflack You're right, must have slipped past when I rebased the branch.

Blazeflack:
How hard would it be to merge this into the ClassIcons element?

p3lim:
Every single classicon is handled by a single texture/frame that all use the same API/system for updates (sans vehicles because of buggy API).
Runes use their own API, and they also have statusbars that needs to be managed and can sometimes proc etc.

In theory it's not hard to merge, but in practicality it's best to leave them separate, same with totems.

Blazeflack:
That makes sense.

p3lim:
Removed the compatibility code since 7.0 is already out.

p3lim:
Removed the compatibility code since 7.0 is already out.

ls-:
Hi!

There's a bug [here](https://github.com/p3lim-wow/oUF/blob/7f2d016a6bdf80179fa49b011032aaa0af9022b2/elements/runebar.lua#L129). Please, replace r, g, b with colors[1], colors[2], colors[3].
```Lua
rune.bg:SetVertexColor(colors[1] * mu, colors[2] * mu, colors[3] * mu)
```

p3lim:
> There's a bug here. Please, replace r, g, b with colors[1], colors[2], colors[3].

Fixed.

PR #272, opened 2016-05-31
Legion combopoints

p3lim:
This PR contains changes to the cpoints element.

These commits are compatible with both live and beta clients, this is not a PR specific to Legion, just something I found while iterating over all the elements while creating the Legion updates.

### DIFF
--- a/elements/cpoints.lua
+++ b/elements/cpoints.lua
@@ -81,7 +81,7 @@ local Enable = function(self)
 		cpoints.__owner = self
 		cpoints.ForceUpdate = ForceUpdate
 
-		self:RegisterEvent('UNIT_COMBO_POINTS', Path, true)
+		self:RegisterEvent('UNIT_POWER_FREQUENT', Path)
 		self:RegisterEvent('PLAYER_TARGET_CHANGED', Path, true)
 
 		for index = 1, MAX_COMBO_POINTS do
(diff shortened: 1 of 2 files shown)


p3lim:
In Legion and post-#266, this element is not really needed any more and could be deprecated.

daneras:
Hi,
i have changed both files elements/tags.lua and cpoints.lua -> the same result. No combo update on the target. Only deselect and targeting again show the actual combos, but no update. Do i make something wrong?
Thx and greetings.

daneras:
Well, sorry, my fault. Combos are shown on the playerframe new ^^

p3lim:
I would highly suggest you try out classicons instead, they cover combo points as well.

daneras:
@p3lim hi, thx for the answer, i will try it, but it works fine now for me.

ls-:
It's Legion now, why dun we just delete CPoints for good? [People are confused](http://www.wowinterface.com/forums/showthread.php?p=316972) and try to do things wrong way.

daneras:
Hi @ls- , yes, everything is fine, thx. It was a mistake by me, i thought the cp were on the target, but they are now on the playerframe. You can close this ticket.
Greetings

ls-:
@daneras, you aren't the only one who asked about CPoints 😁

Issue #177, opened 2013-03-15
Arena preparation frames


haste:
http://www.wowinterface.com/forums/showthread.php?p=274686#post274686

Rainrider:
Closing this in favor of #275

PR #275, opened 2016-07-11
Feature: Arena Preparation

p3lim:
This adds support for arena preparation frames, a solution to #177.

By taking over the `arena#` frames' state driver until the unitIDs become available, this PR achieves the following:
- Displaying the frames with bogus health/power information
- Class color support on health/power elements
- Class color support in `[raidcolor]` tag
- New tag `[arenaspec]` which also work on the normal `arena#` frames

That should cover most of the available features with the arena prep system, and it requires no additional setup for the layout (plug-n-play).

Please test!

### DIFF
--- a/elements/health.lua
+++ b/elements/health.lua
@@ -89,12 +89,19 @@ local oUF = ns.oUF
 oUF.colors.health = {49/255, 207/255, 37/255}
 
 local Update = function(self, event, unit)
-	if(self.unit ~= unit) then return end
+	local arenaPrep = event == 'ArenaPreparation'
+	if(self.unit ~= unit and not arenaPrep) then return end
 	local health = self.Health
 
 	if(health.PreUpdate) then health:PreUpdate(unit) end
 
-	local min, max = UnitHealth(unit), UnitHealthMax(unit)
+	local min, max
+	if(arenaPrep) then
+		min, max = 1, 1
+	else
+		min, max = UnitHealth(unit), UnitHealthMax(unit)
+	end
+
 	local disconnected = not UnitIsConnected(unit)
 	health:SetMinMaxValues(0, max)
 
--- a/elements/health.lua
+++ b/elements/health.lua
@@ -114,7 +114,10 @@ local Update = function(self, event, unit)
 	health.disconnected = disconnected
 
 	local r, g, b, t
-	if(health.colorTapping and not UnitPlayerControlled(unit) and
+	if(health.colorClass and arenaPrep) then
+		local _, _, _, _, _, _, class = GetSpecializationInfoByID(GetArenaOpponentSpec(self.id))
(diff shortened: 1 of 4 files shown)


Sticklord:
I would like request that instead of needing to change all the different elements etc, you could setup the arena frames in a single place.  Because now you need to test if it's an arena preparation frame on all the elements on all units.

We only need to look for the event ARENA_PREP_OPPONENT_SPECIALIZATIONS which enables it and ARENA_OPPONENT_UPDATE which disables it. What I'm proposing is creating function which setups up the dummy frame when we got the information (ARENA_PREP_OPPONENT_SPECIALIZATIONS) and disables it when the arena starts. 

By the way, I tested this and it seemed to work:)

p3lim:
The reason behind the changes to all the elements is because of several reasons:

- The units doesn't exist, so we need to fake them
- When we fake them the elements can't get valid information and would throw errors
- We _do want_ to update the elements to allow the style to be applied

What you are thinking of is a new system to handle this, and that is exactly what we _don't_ want, we want it to be handled gracefully without needing additional work in the layouts.

Sticklord:
Okay, I see your point(s). I just annoying taking care of all the custom elements postupdate in my layout without knowing what event caused it. But it would be nice to have the possibility to handle all this in a single function. Is there an ok way of making the line self:UpdateAllElements('ArenaPreparation', true) instead call a custom setup function? I could always hook it but seems abit needless. The oUF code is abit above my league to understand fully:p 

Or maybe add the classic:

(self.ArenaPreparation or self.UpdateAllElements) (self, 'ArenaPreparation', true)

p3lim:
You could check if the unit exists (`UnitExists(element.__owner.unit)`), I think this should only be nil/false and get to PostUpdate in the occasion of prep updates.

Sticklord:
Okay got it working the way i wanted, thanks!

p3lim:
Need some actual tests on this, I don't do arenas so the only tests I did were while making this feature.
Input from someone that actually does arenas would be nice.

Sticklord:
I tested this, seemed to work nicely. The arena prep frames where on low alpha though (like 50 or so), not sure why that is. 

Those couple of [https://github.com/Sticklord/oUF_Abu/blob/master/LayoutArena.lua#L6](changes) i had was also included but it shouldn't affect it.

Gethe:
I did some testing for this with [my rebased branch](https://github.com/Gethe/oUF/tree/feature-arenaprep). It worked pretty well overall, but I noticed that the frames would not show during arena prep until a teammate enters.

This is honestly a bit of an edge case as it's only really an issue if you're queuing solo for skirmishes, but I figured I would mention it regardless.

p3lim:
@Gethe Might be a bug with the event actually, it worked fine prior to 7.0.

PR #277, opened 2016-07-20
ClassIcons: ClassPowerType color Fix

Manriel:
QuickFix.
Sometimes ClassPowerType is nil and it make a fail.


p3lim:
The correct solution would be to set a fallback power type to `'COMBO_POINTS'.

PR #283, opened 2016-07-22
Legion totems

J4kubS:
Totem categories (fire, water..) have been removed. This fixes issue with ordering where the 1st totem casted is diplayed in the 2nd position, and the 2nd casted totem is displayed in the 1st position.

### DIFF
--- a/elements/totems.lua
+++ b/elements/totems.lua
@@ -74,14 +74,14 @@ local OnLeave = function()
 end
 
 local UpdateTotem = function(self, event, slot)
-	if(slot > MAX_TOTEMS) then return end
 	local totems = self.Totems
+	if(slot > #totems) then return end
 
 	if(totems.PreUpdate) then totems:PreUpdate(slot) end
 
 	local totem = totems[slot]
 	local haveTotem, name, start, duration, icon = GetTotemInfo(slot)
-	if(duration > 0) then
+	if(haveTotem and duration > 0) then
 		if(totem.Icon) then
 			totem.Icon:SetTexture(icon)
 		end
@@ -105,7 +105,7 @@ local Path = function(self, ...)
 end
 
 local Update = function(self, event)
-	for i = 1, MAX_TOTEMS do
+	for i = 1, #self.Totems do
 		Path(self, event, i)
 	end
 end
@@ -121,7 +121,7 @@ local Enable = function(self)
 		totems.__owner = self
 		totems.ForceUpdate = ForceUpdate
 
-		for i = 1, MAX_TOTEMS do
+		for i = 1, #totems do
 			local totem = totems[i]
 
 			totem:SetID(i)
@@ -151,9 +151,11 @@ local Enable = function(self)
 end
 
 local Disable = function(self)
-	if(self.Totems) then
-		for i = 1, MAX_TOTEMS do
-			self.Totems[i]:Hide()
+	local totems = self.Totems
+
+	if(totems) then
+		for i = 1, #totems do
+			totems[i]:Hide()
 		end
 		TotemFrame.Show = nil
 		TotemFrame:Show()
--- a/elements/totems.lua
+++ b/elements/totems.lua
@@ -58,12 +58,6 @@
 local parent, ns = ...
 local oUF = ns.oUF
 
--- Order the list based upon the default UIs priorities.
-local priorities = STANDARD_TOTEM_PRIORITIES
-if(select(2, UnitClass'player') == 'SHAMAN') then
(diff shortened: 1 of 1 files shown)


p3lim:
Can just remove the entire priority logic if you want.

J4kubS:
Updated to remove the whole priority logic. Little bit unsure about the `totems.__map = { unpack(priorities) }` as that might break a layout depending on it.

p3lim:
It's an expansion, breaking changes are to be expected.

Rainrider:
Could you please dump the use of MAX_TOTEMS and just iterate over #totems instead? Layouts usually adapt to changes a lot faster than oUF does and this will give them the ability to do so, at least in this regard. PLAYER_TOTEM_UPDATE handles indices above MAX_TOTEMS fine and currently a shaman can drop up to 5 totems at the same time.

Rainrider:
Also, the TotemFrame is parented to the PlayerFrame, so no need to explicitly hide it.

p3lim:
@Rainrider It's really just an issue on Blizzard's end, MAX_TOTEMS should always be representative as the absolute maximum number of totems. That said, I'll still change it, layouts adapt even faster than Blizzard it seems :unamused:

Rainrider:
MAX_RUNES was the same. Until they removed it. I don't think we should relay on that but rather on the API bahavior. GetTotemInfo returns non-nils up to index 5. MAX_TOTEMS is just a lua global, and Blizzard lags on the UI a lot.

J4kubS:
How about adding a `.num` options like the one in auras? It could default to either `MAX_TOTEMS` or `5` and it would also give more control to the layout.

p3lim:
The table size would always be there so it's the better option.

Rainrider:
What I mean is something like that:
```lua
local UpdateTotem = function(self, event, slot)
	local totems = self.Totems
	local totem = totems[slot]
	if(not totem) then return end -- or if appropriate assert(totem, "Totem update fired for a non-present widget.")

	-- leave rest unchanged
end

local Update = function(self, event)
	for i = 1, #self.Totems do
		Path(self, event, i)
	end
end
```

J4kubS:
I kept the `>` check because sparse arrays would cause trouble in the Update function. To support sparse arrays (skipping certain totem slots) we should use some constant instead of the array size.

Rainrider:
You need the nil check because GetTotemInfo returns information only for the first 5 totem slots. If it is a valid totem slot but no totem for that slot is placed the returns are like:
>Dump: value=GetTotemInfo(1)
>[1]=false,
>[2]="",
>[3]=0,
>[4]=0,
>[5]=""

However:
>Dump: value=GetTotemInfo(6)
>empty result

So in case Blizzard reduces the number of valid slots to say 4 again and a layout still provides 5 widgets, UpdateTotem will error with `attempt to compare number with nil`.

p3lim:
The idea is that the elements should perform basic operations, basically mimic whatever the behavior of their stock UI counterparts, and it's why we have overrides should the layout want a different behavior.

Something like sparse updates could be done with overrides.

J4kubS:
Yes, sorry, I meant the `if(not totem) then return end` check in your example. You're correct about the `if(haveTotem and duration > 0)` which I've missed.

Rainrider:
Not sure how `if(not totem) then return end` relates to sparse arrays. I put it there since it is used 2 lines later either way. The way you did it is absolutely fine too.

elements/totems.lua
```diff
@@ -74,14 +74,14 @@ local OnLeave = function()
...
-	if(slot > MAX_TOTEMS) then return end
 	local totems = self.Totems
+	if(slot > #totems) then return end
 
 	if(totems.PreUpdate) then totems:PreUpdate(slot) end
 
 	local totem = totems[slot]
 	local haveTotem, name, start, duration, icon = GetTotemInfo(slot)
-	if(duration > 0) then
+	if(hateTotem and duration > 0) then
```
p3lim: Typo here

J4kubS:
You could have a Totems array with `Totems[3] = nil` - the `UpdateTotem` would keep working, but the `Update` would update just the first 2. The `if(slot > #totems) then return end` keeps it consistent.

Rainrider:
It will fail in Enable in that case - `totem:SetID(i)` will error with `attempt to index local 'totem' (a nil value)`. This was never supported and using the `.Override` option won't be enough to work around it. The layout can use `totem.Show = totem.Hide` in that case though, or `totem:Hide()` in PostUpdate.

PR #285, opened 2016-07-26
power: Alternative power colours use both 0-1 and 0-255 ranges.

ls-:
Hi!

Title says it all. Here's an example.

image

### DIFF
--- a/elements/power.lua
+++ b/elements/power.lua
@@ -192,10 +192,10 @@ local Update = function(self, event, unit)
 			if(power.GetAlternativeColor) then
 				r, g, b = power:GetAlternativeColor(unit, ptype, ptoken, altR, altG, altB)
 			elseif(altR) then
-				if(altR >= 0 and altR <= 1) and (altG >= 0 and altG <= 1) and (altB >= 0 and altB <= 1) then
-					r, g, b = altR, altG, altB
-				else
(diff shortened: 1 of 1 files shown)


ls-:
Ugh, it just seems to be a simple inconsistency in return values.
image

Blazeflack:
Gotta love inconsistency like that.

p3lim:
Where can you test this?

ls-:
@p3lim Iron Reaver in HFC (first screenshot), and Beth'tilac in FL (second screenshot).

ls-:
K, there's is an issue with {1, 1 ,1} cases. cuz I excluded 1s.

I'll change it, when I get home...
```Lua
if(altR > 0 and altR <= 1) and (altG > 0 and altG <= 1) and (altB > 0 and altB <= 1) then```

Rainrider:
`if (altR > 1) or (altG > 1) or (altG > 1) then r, g, b = altR/255, ...` is more readable, I think

ls-:
@Rainrider, fuck, logic failed me, ahahaha 😂 

Will update it later.

p3lim:
Seems like a bug with the API, you should add a comment to that block that denotes that (like https://github.com/haste/oUF/commit/1bde1993b85985471878bc7392d6d282ea6590e0)

PR #286, opened 2016-07-26
power: blizzard's atlas for insanity, maelstorm, lunar power bars

bbeny123:
Also updates data from FrameXML/Constants.lua required to proper working of this feature

### DIFF
--- a/elements/power.lua
+++ b/elements/power.lua
@@ -61,8 +61,9 @@
 
  Examples
 
-   -- Position and size
+   -- Position, size and custom layout
    local Power = CreateFrame("StatusBar", nil, self)
+   Power.statusbar = "StatusBar"
    Power:SetHeight(20)
    Power:SetPoint('BOTTOM')
    Power:SetPoint('LEFT')
@@ -80,6 +81,7 @@
    Power.colorPower = true
    Power.colorClass = true
    Power.colorReaction = true
+   Power.atlasEnable = true
    
    -- Make the background darker.
    Background.multiplier = .5
@@ -99,6 +101,8 @@ local parent, ns = ...
 local oUF = ns.oUF
 
 local isBetaClient = select(4, GetBuildInfo()) >= 70000
+-- required to proper working SetStatusBarAtlas()
+local statusbar = [[Interface\TargetingFrame\UI-StatusBar]] -- default StatusBarTexture
 
 oUF.colors.power = {}
 for power, color in next, PowerBarColor do
@@ -110,7 +114,7 @@ for power, color in next, PowerBarColor do
 				oUF.colors.power[power][index] = {color.r, color.g, color.b}
 			end
 		else
-			oUF.colors.power[power] = {color.r, color.g, color.b}
+			oUF.colors.power[power] = {color.r, color.g, color.b, atlas = color.atlas}
 		end
 	end
 end
@@ -208,9 +212,23 @@ local Update = function(self, event, unit)
         local adjust = 0 - (min or 0)
 		r, g, b = self.ColorGradient(cur + adjust, max + adjust, unpack(power.smoothGradient or self.colors.smooth))
 	end
-
+	
+	if(power.statusbar) then
+		statusbar = power.statusbar
+	end
+	
 	if(t) then
-		r, g, b = t[1], t[2], t[3]
+		if ( power.colorPower and power.atlasEnable ) then
+			if ( t.atlas ) then
+				power:SetStatusBarAtlas(t.atlas);
+				power:SetStatusBarColor(1, 1, 1);
+			else
+				power:SetStatusBarTexture(statusbar)
+				r, g, b = t[1], t[2], t[3]
+			end
+		else
+			r, g, b = t[1], t[2], t[3]
+		end
 	end
 
 	if(b) then
@@ -258,7 +276,10 @@ local Enable = function(self, unit)
 		self:RegisterEvent('UNIT_FACTION', Path)
 
 		if(power:IsObjectType'StatusBar' and not power:GetStatusBarTexture()) then
-			power:SetStatusBarTexture[[Interface\TargetingFrame\UI-StatusBar]]
+			if(power.colorPower and power.statusbar and power.atlasEnable) then
+				statusbar = power.statusbar
+			end
+			power:SetStatusBarTexture(statusbar)
 		end
 
 		return true


bbeny123:
changes

ls-:
Hi!

There's one big issue w/ this PR, it's quite dirty. You should base it on [master's version of power](https://github.com/haste/oUF/blob/master/elements/power.lua) without code from other patches/PRs. I can see code from my power fix PR, p3lim's arena prep frames, also try not to include things that aren't related to what this PR is supposed to do, like do not clean trailing whitespaces, etc.

ls-:
K. one issue [here ](https://github.com/bbeny123/oUF/blob/15597b29e31ecafb19f7c736b612b96f8fd8ae66/elements/power.lua#L221-L230), if there a status bar texture set in layout, it shouldn't happen at all.

bbeny123:
Hi

> There's one big issue w/ this PR, it's quite dirty. You should base it on master's version of power without code from other patches/PRs 

Sorry for that, initially i forked your rep not haste's. I already changed this, and now this pull request added only legion style power bars :)

p3lim:
Please keep your PR clean, don't include changes that are irrelevant to the feature you're promoting, and once you're done committing, please rebase the commit(s) to as few as needed, formatting the commit message as such: "elementname: changes".

As far as the feature goes, it's a neat idea, but how it's enabled needs to be changed to be optional, currently it's considered opt-out.

bbeny123:
Now should be ok, its fully optional and i merged all commits to one :)

Blazeflack:
The commit still removes beta checks. These changes are irrelevant to this feature and should not be included.

p3lim:
It should also be tied to the `colorPower` option imo.

bbeny123:
> The commit still removes beta checks. These changes are irrelevant to this feature and should not be included.

fixed

> It should also be tied to the colorPower option imo.

added

Blazeflack:
>fixed

Not quite yet. You're still making changes that are unrelated to this feature.

See this:
http://i.imgur.com/6ohNcc7.png

bbeny123:
This changes are related, because this feature using a data from FrameXML/Constants.lua. 
i.e. this feature wouldn't work for chi, if chi would have atlas (without this changes)

p3lim:
- Chi has always been been on index 12, even back when it was called "Light Force" (in MoP alpha they were toying with the idea of having light and dark chi, a yin-yan theme as their resource).
- Combopoints has always been on index 4.
- Arcane charges didn't exist in WoD

[source](https://www.townlong-yak.com/framexml/beta/Constants.lua/diff#108)

Regardless, even if it _was_ wrong, it would be a bug and required a different PR.

bbeny123:
> Regardless, even if it was wrong, it would be a bug and required a different PR.

Ok, i revetred this change. I will make a different PR for it.

ls-:
@p3lim, there's some inconsistency on Blizz side, yet again, [here](https://www.townlong-yak.com/framexml/beta/UnitFrame.lua#30), you can see that Chi sits at 4 T_T

p3lim:
@ls- The reason we follow the ones in Constants.lua is because we're using those constants.

ls-:
@p3lim, yeah, Ikr T_T I just wanted to show, that there's inconsistency and person probably was confused by it.

gagou:
I think that we should have an option to use the power atlas for the bg if it exists.


Currently bg color is not set correctly when using an atlas as r, g, b are not given values.

Instead of doing:
`power:SetStatusBarColor(1, 1, 1)`

I would do:
`r, g, b = 1, 1, 1`

elements/power.lua
```diff
@@ -258,7 +276,10 @@ local Enable = function(self, unit)
 		self:RegisterEvent('UNIT_FACTION', Path)
 
 		if(power:IsObjectType'StatusBar' and not power:GetStatusBarTexture()) then
-			power:SetStatusBarTexture[[Interface\TargetingFrame\UI-StatusBar]]
+			if(power.colorPower and power.statusbar and power.atlasEnable) then
+				statusbar = power.statusbar
```
p3lim: The layout should not be required to add extra information for internal logic, a better way of handling this is to set the original statusbar variable _after_ the fallback code, and just use normal methods of getting it.

Basically, after [this line](https://github.com/haste/oUF/blob/441ee4be0d1140f7861e416609760cf7003a36be/elements/power.lua#L262) you could do `statusbar = power:GetStatusBarTexture()` (but you also have to consider the posibility that the object might not be a statusbar, requiring a teensy bit more logic).

Also, different statusbars might have different textures, storing them globally for the entire element is not a good idea, you should store it on the element itself (e.g. `power._originalTexture`).

elements/power.lua
```diff
@@ -208,9 +212,23 @@ local Update = function(self, event, unit)
         local adjust = 0 - (min or 0)
 		r, g, b = self.ColorGradient(cur + adjust, max + adjust, unpack(power.smoothGradient or self.colors.smooth))
 	end
-
+	
+	if(power.statusbar) then
```
p3lim: What is the purpose of this? It's basically a duplicate of line 280.

elements/power.lua
```diff
@@ -208,9 +212,23 @@ local Update = function(self, event, unit)
...
+			else
+				power:SetStatusBarTexture(statusbar)
+				r, g, b = t[1], t[2], t[3]
+			end
+		else
+			r, g, b = t[1], t[2], t[3]
+		end
 	end
 
 	if(b) then
```
p3lim: The variables `r`, `g` and `b` can still exist here, even if atlas is used, and will result in a discolored bar.

elements/power.lua
```diff
@@ -208,9 +212,23 @@ local Update = function(self, event, unit)
         local adjust = 0 - (min or 0)
 		r, g, b = self.ColorGradient(cur + adjust, max + adjust, unpack(power.smoothGradient or self.colors.smooth))
 	end
-
+	
+	if(power.statusbar) then
+		statusbar = power.statusbar
+	end
+
```
p3lim: What is the purpose of this block?

elements/power.lua
```diff
@@ -208,9 +212,23 @@ local Update = function(self, event, unit)
         local adjust = 0 - (min or 0)
 		r, g, b = self.ColorGradient(cur + adjust, max + adjust, unpack(power.smoothGradient or self.colors.smooth))
 	end
-
+	
+	if(power.statusbar) then
+		statusbar = power.statusbar
+	end
```
p3lim: What is the purpose of this block?

elements/power.lua
p3lim: The variables r, g and b can still exist even if atlas is used, and will result in a discolored bar (although the situation might be rare, it's best to have a failsafe).

elements/power.lua
p3lim: The layout should not be required to add extra information for internal logic, a better way of handling this is to set the original statusbar variable after the fallback code, and just use normal methods of getting it.

Basically, after [this line](https://github.com/oUF-wow/oUF/blob/441ee4be0d1140f7861e416609760cf7003a36be/elements/power.lua#L262) you could do statusbar = power:GetStatusBarTexture() (but you also have to consider the posibility that the object might not be a statusbar, requiring a teensy bit more logic).

Also, different statusbars might have different textures, storing them globally for the entire element is not a good idea, you should store it on the element itself (e.g. power._originalTexture).

ls-:
Closing this one in favour of #304.

@bbeny123, thanks for the idea! 👍

PR #287, opened 2016-07-29
power: changed and updated source for colors

bbeny123:
As a source for oUF.colors.power should be used FrameXML/UnitFrame.lua not the FrameXML/Constants.lua. Why? Because [here](https://github.com/haste/oUF/blob/master/elements/power.lua#L104) is used a PowerBarColor function, which gets data from FrameXML/UnitFrame.lua

proof:
`/run for k, v in pairs( PowerBarColor[4]  ) do print(k, v) end`
After runinig this macro in WoW we get: 
gskaep2
This is color of Chi not the Combo Points, so old source is wrong.

### DIFF
--- a/elements/power.lua
+++ b/elements/power.lua
@@ -98,8 +98,6 @@
 local parent, ns = ...
 local oUF = ns.oUF
 
-local isBetaClient = select(4, GetBuildInfo()) >= 70000
-
 oUF.colors.power = {}
 for power, color in next, PowerBarColor do
 	if (type(power) == "string") then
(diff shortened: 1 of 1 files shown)


bbeny123:
link to Legion [UnitFrame.lua ](https://www.townlong-yak.com/framexml/beta/UnitFrame.lua)

p3lim:
We use the strings of `PowerBarColor` for the actual colors (and atlases), defined in [FrameXML/UnitFrame.lua](https://github.com/Gethe/wow-ui-source/blob/beta/FrameXML/UnitFrame.lua#L2-L22), but we use the indices defined in [FrameXML/Constants.lua](https://github.com/Gethe/wow-ui-source/blob/beta/FrameXML/Constants.lua#L108-L127), because of two reasons:

1. The ones defined in `FrameXML/Constants.lua` are correct.
2. Even if they _were_ incorrect (which they're not), we're using the constants from that file and we need things to match up.

You can test yourself, play on a windwalker monk, get some Chi and run this:
```Lua
print(UnitPower('player', 4)) -- This is combo points
print(UnitPower('player', 12)) -- This is chi
```

PR #298, opened 2016-09-13
core:  update the pet frame properly after entering/exiting a vehicle

Rainrider:
Fixes https://github.com/oUF-wow/oUF/issues/257
Closes https://github.com/oUF-wow/oUF/pull/258

It would be nice if someone manages to test this is a raid/party

### DIFF
--- a/ouf.lua
+++ b/ouf.lua
@@ -46,12 +46,7 @@ local updateActiveUnit = function(self, event, unit)
 		realUnit = 'target'
 	end
 
-	if(modUnit == "pet" and realUnit ~= "pet") then
-		modUnit = "vehicle"
-	end
-
-	-- Drop out if the event unit doesn't match any of the frame units.
-	if(not UnitExists(modUnit) or unit and unit ~= realUnit and unit ~= modUnit) then return end
+	if(not UnitExists(modUnit)) then return end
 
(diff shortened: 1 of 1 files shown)


gagou:
I've done some test with my hunter in trial of the crusader and did not see any problem

Rainrider:
@gagou we're you in a group? I mean I tested the player and pet frames myself. The interesting part is whether party/raid members are updated properly.

ls-:
@gagou, we're talking about testing raid#, raid#pet, group#, group#pet frames now, cuz they might or might not be buggy 😕

gagou:
I tested it in a party in trial of crusader with my hunt and mage (I have 3 wow accounts)
Group and grouppet frames are correctly updating when going in and out of vehicle, same for player and pet frame.
Raid frames are correctly updating, I tested them in Ulduar.
Can't test Raid pet I don't display pet for raid members in my layout.

Rainrider:
@gagou Which layout did you use for testing? Or are you using headers for party/raid frames and the corresponding pets if you know that?

gagou:
It's my own layout, I'm using header for party / raid

Rainrider:
OK, thank you very much

PR #300, opened 2016-09-17
totems: TotemFrame is parented to PlayerFrame

p3lim:
It will be hidden with oUF:DisableBlizzard()

Mentioned here: https://github.com/oUF-wow/oUF/pull/283#issuecomment-239676805


p3lim:
Was not sure if I wanted to kill the registering/unregistering of the TotemFrame events or not.

ls-:
It's fine, better safe than sorry...

PR #303, opened 2016-09-22
core: Arena and Boss frames' statusbars have no key reference

p3lim:


p3lim:
Nevermind, they get their key reference after being created.

PR #304, opened 2016-09-23
power: allow using atlases

Rainrider:
Another attempt on #286 

~~**Issues:**~~
~~I was not able to unset the atlas once it was set, despite following [blizzard's code](https://github.com/Gethe/wow-ui-source/blob/live/FrameXML/UnitFrame.lua#L486-L498). Also tried `power:SetStatusBarAtlas(nil)` which made `power:GetStatusBarAtlas()` return nil but it didn't change anything (btw `power:SetStatusBarAtlas()` errors, nil is required like in `:SetTexture`). So I must be missing something here, but don't know what.~~
~~To reproduce:~~
  1. ~~set `Power.useAtlas` to true for the target frame in your layout~~
  2. ~~create a demon hunter~~
  3. ~~target yourself and then switch target to a normal mob (with a power resource not using an atlas)~~

**Limitations:**
All the coloring options in the power element could not be applied to atlases as the results depend on the color of the atlas.

**Notes:**
Atlases could be actually handled by the layout. To better support that we could just pass `ptoken` and `ptype` to PostUpdate and copy atlas information to the oUF.colors.power. ~~However the issue with unsetting the atlas remains.~~

### DIFF
--- a/elements/power.lua
+++ b/elements/power.lua
@@ -71,12 +71,12 @@
    Power:SetPoint('BOTTOM')
    Power:SetPoint('LEFT')
    Power:SetPoint('RIGHT')
-   
+
    -- Add a background
    local Background = Power:CreateTexture(nil, 'BACKGROUND')
    Background:SetAllPoints(Power)
    Background:SetTexture(1, 1, 1, .5)
-   
+
    -- Options
    Power.frequentUpdates = true
    Power.colorTapping = true
@@ -84,10 +84,10 @@
    Power.colorPower = true
    Power.colorClass = true
    Power.colorReaction = true
-   
+
    -- Make the background darker.
    Background.multiplier = .5
-   
+
    -- Register it with oUF
    self.Power = Power
    self.Power.bg = Background
--- a/elements/power.lua
+++ b/elements/power.lua
@@ -208,7 +208,7 @@ local Update = function(self, event, unit)
 	end
 
 	t = self.colors.power[ptoken or ptype]
-	local atlas = power.atlas or t.atlas
+	local atlas = power.atlas or (t and t.atlas)
 	if(power.useAtlas and atlas and displayType ~= ALTERNATE_POWER_INDEX) then
 		power:SetStatusBarAtlas(atlas)
 		power:SetStatusBarColor(1, 1, 1)
@@ -216,7 +216,7 @@ local Update = function(self, event, unit)
 			t = disconnected and self.colors.disconnected or self.colors.tapped
 			power:GetStatusBarTexture():SetDesaturated(disconnected or tapped)
 		end
-		if(b) then
+		if(t and b) then
 			r, g, b = t[1], t[2], t[3]
 		end
 	else
--- a/elements/power.lua
+++ b/elements/power.lua
@@ -21,6 +21,10 @@
  .displayAltPower   - Use this to let the widget display alternate power if the
                       unit has one. If no alternate power the display will fall
                       back to primary power.
+ .useAtlas          - Use this to let the widget use an atlas for its texture if
+                      `.atlas` is defined on the widget or an atlas is present in
+                      `self.colors.power` for the appropriate power type.
+ .atlas             - A custom atlas
 
  The following options are listed by priority. The first check that returns
  true decides the color of the bar.
@@ -108,7 +112,7 @@ for power, color in next, PowerBarColor do
 				oUF.colors.power[power][index] = {color.r, color.g, color.b}
 			end
 		else
-			oUF.colors.power[power] = {color.r, color.g, color.b}
+			oUF.colors.power[power] = {color.r, color.g, color.b, atlas = color.atlas}
 		end
 	end
 end
@@ -150,6 +154,7 @@ local Update = function(self, event, unit)
 	end
 	local cur, max = UnitPower(unit, displayType), UnitPowerMax(unit, displayType)
 	local disconnected = not UnitIsConnected(unit)
+	local tapped = not UnitPlayerControlled(unit) and UnitIsTapDenied(unit)
 	power:SetMinMaxValues(min or 0, max)
 
 	if(disconnected) then
@@ -159,7 +164,9 @@ local Update = function(self, event, unit)
 	end
 
 	power.disconnected = disconnected
+	power.tapped = tapped
 
+	local ptype, ptoken, altR, altG, altB = UnitPowerType(unit)
 	local r, g, b, t
 
 	if(power.colorTapping and not UnitPlayerControlled(unit) and UnitIsTapDenied(unit)) then
@@ -169,8 +176,6 @@ local Update = function(self, event, unit)
 	elseif(displayType == ALTERNATE_POWER_INDEX and power.altPowerColor) then
 		t = power.altPowerColor
 	elseif(power.colorPower) then
-		local ptype, ptoken, altR, altG, altB = UnitPowerType(unit)
-
 		t = self.colors.power[ptoken]
 		if(not t) then
 			if(power.GetAlternativeColor) then
@@ -194,7 +199,7 @@ local Update = function(self, event, unit)
 	elseif(power.colorReaction and UnitReaction(unit, 'player')) then
 		t = self.colors.reaction[UnitReaction(unit, "player")]
 	elseif(power.colorSmooth) then
-        local adjust = 0 - (min or 0)
+		local adjust = 0 - (min or 0)
 		r, g, b = self.ColorGradient(cur + adjust, max + adjust, unpack(power.smoothGradient or self.colors.smooth))
 	end
 
@@ -202,18 +207,33 @@ local Update = function(self, event, unit)
 		r, g, b = t[1], t[2], t[3]
 	end
 
-	if(b) then
-		power:SetStatusBarColor(r, g, b)
-
-		local bg = power.bg
-		if(bg) then
-			local mu = bg.multiplier or 1
-			bg:SetVertexColor(r * mu, g * mu, b * mu)
+	t = self.colors.power[ptoken or ptype]
+	local atlas = power.atlas or t.atlas
+	if(power.useAtlas and atlas and displayType ~= ALTERNATE_POWER_INDEX) then
+		power:SetStatusBarAtlas(atlas)
+		power:SetStatusBarColor(1, 1, 1)
+		if(power.colorTapping or power.colorDisconnected) then
+			t = disconnected and self.colors.disconnected or self.colors.tapped
+			power:GetStatusBarTexture():SetDesaturated(disconnected or tapped)
+		end
+		if(b) then
+			r, g, b = t[1], t[2], t[3]
+		end
+	else
+		power:SetStatusBarTexture(power.texture)
+		if(b) then
+			power:SetStatusBarColor(r, g, b)
 		end
 	end
 
+	local bg = power.bg
+	if(bg and b) then
+		local mu = bg.multiplier or 1
+		bg:SetVertexColor(r * mu, g * mu, b * mu)
+	end
+
 	if(power.PostUpdate) then
-		return power:PostUpdate(unit, cur, max, min)
+		return power:PostUpdate(unit, cur, max, min, ptoken, ptype)
 	end
 end
 
@@ -246,8 +266,9 @@ local Enable = function(self, unit)
 		-- For tapping.
 		self:RegisterEvent('UNIT_FACTION', Path)
 
-		if(power:IsObjectType'StatusBar' and not power:GetStatusBarTexture()) then
-			power:SetStatusBarTexture[[Interface\TargetingFrame\UI-StatusBar]]
+		if(power:IsObjectType'StatusBar') then
+			power.texture = power:GetStatusBarTexture() and power:GetStatusBarTexture():GetTexture() or [[Interface\TargetingFrame\UI-StatusBar]]
+			power:SetStatusBarTexture(power.texture)
 		end
 
 		return true


ls-:
IIRC, you don't really need to unset the atlas, cuz :SetTexture and :SetAtlas are mutually exclusive...

Rainrider:
> P.S. Tested, works fine 👍

You mean you didn't have the issue from above?

Rainrider:
I force pushed a change to update the background texture too if present. Maybe that was causing the issue for me - it is now gone.

If you like it as it is I will update the documentation and then it could be merged. Or we offload it to the layout, which would keep the element code simpler.

elements/power.lua
```diff
@@ -160,58 +161,72 @@ local Update = function(self, event, unit)
...
 			end
+		elseif(power.colorClass and UnitIsPlayer(unit)) or
+			(power.colorClassNPC and not UnitIsPlayer(unit)) or
+			(power.colorClassPet and UnitPlayerControlled(unit) and not UnitIsPlayer(unit)) then
+			local _, class = UnitClass(unit)
+			t = self.colors.class[class]
+		elseif(power.colorReaction and UnitReaction(unit, 'player')) then
+			t = self.colors.reaction[UnitReaction(unit, "player")]
+		elseif(power.colorSmooth) then
+	        local adjust = 0 - (min or 0)
```
p3lim: Mixed indenting

elements/power.lua
```diff
@@ -160,58 +161,72 @@ local Update = function(self, event, unit)
...
+	local ptype, ptoken, altR, altG, altB = UnitPowerType(unit)
+	local t = self.colors.power[ptoken]
+	local r, g, b
+	local atlas = power.atlas or t.atlas
+	local bg = power.bg
+
+	if(power.useAtlas and atlas and displayType ~= ALTERNATE_POWER_INDEX) then
+		r, g, b = t[1], t[2], t[3]
+		power:SetStatusBarAtlas(atlas)
+		power:SetStatusBarColor(1, 1, 1)
```
p3lim: Is this necessary? Haven't tested this myself, just wondering.

elements/power.lua
```diff
@@ -246,8 +261,13 @@ local Enable = function(self, unit)
...
 
-		if(power:IsObjectType'StatusBar' and not power:GetStatusBarTexture()) then
-			power:SetStatusBarTexture[[Interface\TargetingFrame\UI-StatusBar]]
+		if(power:IsObjectType'StatusBar') then
+			power.texture = power:GetStatusBarTexture() and power:GetStatusBarTexture():GetTexture() or [[Interface\TargetingFrame\UI-StatusBar]]
+			power:SetStatusBarTexture(power.texture)
+		end
+		
+		if(power.bg) then
+			power.bg.texture = power.bg:GetTexture()
```
p3lim: Backgrounds don't always contain a texture, can be a texture color.

elements/power.lua
```diff
@@ -160,58 +161,72 @@ local Update = function(self, event, unit)
...
+	local bg = power.bg
+
+	if(power.useAtlas and atlas and displayType ~= ALTERNATE_POWER_INDEX) then
+		r, g, b = t[1], t[2], t[3]
+		power:SetStatusBarAtlas(atlas)
+		power:SetStatusBarColor(1, 1, 1)
+		power:GetStatusBarTexture():SetDesaturated(disconnected or tapped)
+
+		if(bg) then
+			bg:SetAtlas(atlas)
```
p3lim: This conflicts with the following block at lines 225-228:
```Lua
if(bg) then
	local mu = bg.multiplier or 1
	bg:SetVertexColor(r * mu, g * mu, b * mu)
end
```

elements/power.lua
ls-: Yeah, you can change atlas' colour, cuz it's a texture. However, instead of `t[1], ...` values, I'd use `0.5`, this way bg texture would look like a darkened version of fg (atlas). 

TBH, I wouldn't set atlas for bg texture at all. There might be some fancy texture, some half-transparent texture, or even half-transparent colour. I'd simply set its vertex colour to 0.5s.

elements/power.lua
ls-: It doesn't. Atlas is a texture, and bg should differ from fg.

elements/power.lua
Rainrider: Actually I've tried not using atlas for the background and I use 0.5 for the multiplier. It was hard to see the power bar fill with some atlases. Because of that I would rather let the layout decide on the multiplier value since they could react if the contrast is not good enough.

I liked the atlas on the background actually, I could make it optional though.

I would rather not copy coloring code to the atlas section. If the layout prefers to use an atlas I suppose they like the default texture colors. They could also supply a custom atlas through `power.atlas`, alter `oUF.colors.power` or use PostUpdate if they need coloring for some reason.

elements/power.lua
Rainrider: I made an upsi here 😊

elements/power.lua
Rainrider: Valid. Will look into it.

p3lim: Should probably be looked into in other elements, but that's for another time (and PR).

elements/power.lua
p3lim: :+1:

elements/power.lua
Rainrider: `:SetColorTexture(1, 1, 1)` would make `:GetTexture()` return "Color-ffffffff"
`:SetColorTexture(1, 0, 0)` -> "Color-ff00ff"
`:SetColorTexture(0.5, 0, 0)` -> "Color-8000ff"
`SetColorTexture(0.6, 0, 0)` -> "Color-9900ff"
`SetColorTexture(0.01, 0, 0)` -> "Color-300ff"

It will be challenging to extract the color from that

elements/power.lua
Rainrider: Oh, got you wrong on the copying thing. You mean what if ptoken is nil, right? If `power.atlas` is non-nil, it will error. Could use ptype as a fail-safe then, don't know if this is optimal though.

ls-: Hope we aren't talking about `bg.multiplier` now 😁 

Anyways, here's my test setup:
```Lua
self:SetSize(238, 24)

local power = CreateFrame("StatusBar", nil, self)
power:SetAllPoints()
power:SetStatusBarTexture("Interface\\BUTTONS\\WHITE8X8")
power.colorPower = true
power.frequentUpdates = true
-- power.useAtlas = true

local bg = power:CreateTexture(nil, 'BACKGROUND')
bg:SetAllPoints(power)
bg:SetColorTexture(1, 1, 1, .4)
power.bg = bg

self.Power = power
```
Your current implementation breaks it at [#L118](https://github.com/oUF-wow/oUF/blob/cac1dec131163035591751ee3be2075e062ccb5c/elements/power.lua#L181), cuz `power.bg.texture` is `Color-ffffff80`, so I commented out that bg texture reset.

That's how it looks like:

image

Then I enabled atlas:
image

As you can see, cuz `bg.multiplier` is 1, there's nearly no difference between fg and bg textures.

Then I removed `bg:SetAtlas(atlas)` thingy:
image

Better, and you don't need to worry about bg being a texture or color-texture 😁 But `t` may be `nil`, that's why I suggested copying that block from [#L190-L203](https://github.com/oUF-wow/oUF/blob/cac1dec131163035591751ee3be2075e062ccb5c/elements/power.lua#L190-L203). But code duplication sucks, mb reimplement that block as `ResolvePowerColor` function? 😒 

Or you could replace `r, g, b = t[1], t[2], t[3]` with `r, g, b = 0, 0, 0`:
image

This way you don't need to worry about anything.

elements/power.lua
```diff
@@ -160,33 +164,30 @@ local Update = function(self, event, unit)
 	end
 
 	power.disconnected = disconnected
-
+	power.tapped = tapped
+	
 	local ptype, ptoken, altR, altG, altB = UnitPowerType(unit)
-	local t = self.colors.power[ptoken]
+	local t = self.colors.power[ptoken or ptype or 'MANA']
```
ls-: TBH, I'd leave it as `local t = self.colors.power[ptoken]`

Rainrider:
I hope I addressed everything. 

  - No more atlas for bg because of how textures work when `:SetColorTexture` is used.
  - Updated the documentation.
  - Use `ptype` or `'MANA'` as a fail-safe for the color for atlases.
  - Honor `.colorDisconnected` and `.colorTapped` before desaturating the status bar for atlases.
  - Pass `ptoken` and `ptype` to PostUpdate.
  - Add `.tapped` to the widget.
  - Fix mixed indentation.

Further, this is what I use in PostUpdate since I like atlas for bg:
```lua
local atlas = Power:GetStatusBarAtlas()
if (atlas) then
	Power.bg:SetAtlas(atlas)
	Power.bg:SetDesaturated(Power.disconnected or Power.tapped)
else
	Power.bg:SetTexture(ns.media.TEXTURE)
end
```

If you approve, I'll squash both commits and merge in master.

elements/power.lua
```diff
@@ -160,33 +164,30 @@ local Update = function(self, event, unit)
...
-		if(bg) then bg:SetTexture(bg.texture) end
-
 		if(power.colorTapping and tapped) then
 			t = self.colors.tapped
 		elseif(power.colorDisconnected and disconnected) then
 			t = self.colors.disconnected
 		elseif(displayType == ALTERNATE_POWER_INDEX and power.altPowerColor) then
 			t = power.altPowerColor
 		elseif(power.colorPower) then
+			t = self.colors.power[ptoken]
```
ls-: and remove this line here.

IMHO, It'd look better and more consistent this way😁 But that's just my preference, you may ignore it.

elements/power.lua
```diff
@@ -160,33 +164,30 @@ local Update = function(self, event, unit)
...
 	local ptype, ptoken, altR, altG, altB = UnitPowerType(unit)
-	local t = self.colors.power[ptoken]
+	local t = self.colors.power[ptoken or ptype or 'MANA']
 	local r, g, b
 	local atlas = power.atlas or t.atlas
-	local bg = power.bg
-
+	
 	if(power.useAtlas and atlas and displayType ~= ALTERNATE_POWER_INDEX) then
 		r, g, b = t[1], t[2], t[3]
```
ls-: but add `t = t or self.colors.power[ptype or 'MANA']` here

Sticklord:
There's a fault here, if no color option is set for the bar, it will fallback to this:

local t = self.colors.power[ptoken or ptype or 'MANA']

Which mean you can't set a custom color for the bar without Power:Override() or some other hack.

ls-:
@Sticklord, just to make things clear and avoid confusion. Are you talking about `.colorClass`, etc options now?

Sticklord:
Yeah those, it was how the power element was working before. For example inserting `t = nil` on line 182 would fix this.

ls-:
K. I guess, there's some kind of misunderstanding here. In this case we recolour only `bg` texture, and, IMHO, it should be coloured according to power type, cuz atlases are predefined textures, and their colour always represents power type.

elements/power.lua
```diff
@@ -159,61 +164,73 @@ local Update = function(self, event, unit)
...
-			if(power.GetAlternativeColor) then
-				r, g, b = power:GetAlternativeColor(unit, ptype, ptoken, altR, altG, altB)
-			elseif(altR) then
-				-- As of 7.0.3, altR, altG, altB may be in 0-1 or 0-255 range.
-				if(altR > 1) or (altG > 1) or (altB > 1) then
-					r, g, b = altR / 255, altG / 255, altB / 255
+	power.tapped = tapped
+	
+	local ptype, ptoken, altR, altG, altB = UnitPowerType(unit)
+	local t = self.colors.power[ptoken or ptype or 'MANA']
```
Sticklord: Before, "t" did not get assigned any value.

Sticklord: You set t to a color before the checks which makes certain options invalid.

elements/power.lua
```diff
@@ -159,61 +164,73 @@ local Update = function(self, event, unit)
...
-		r, g, b = self.ColorGradient(cur + adjust, max + adjust, unpack(power.smoothGradient or self.colors.smooth))
-	end
 
-	if(t) then
-		r, g, b = t[1], t[2], t[3]
-	end
-
-	if(b) then
-		power:SetStatusBarColor(r, g, b)
+		if(t) then
```
Sticklord: Because t already exist.

Sticklord:
Sorry it's a mess, first time using that, but hope you understand.

ls-:
@Sticklord, haha, its fine, I already requested said changes [here](https://github.com/oUF-wow/oUF/pull/304#pullrequestreview-1435125) 😁 

But you gave us another reason to fix that part.

Sticklord:
Oh shit, yeah guess I'm the one that need sleep the most:P

ls-:
I updated my review, and included your suggestions 😁 

P.S. Good night, I need to sleep badly 😪

elements/power.lua
Rainrider: The order of preference for the color if `.colorPower` is set is:
  1. `ptoken`
  2. `power.GetAlternativeColor`
  3. `altR, altG, altB`
  4. `ptype`

The order of preference for atlas is `ptoken`, `ptype`, or `'MANA'`.

Because of this I had to reset the color table here, I wanted to keep the current behavior for that option.

elements/power.lua
Rainrider: Or change line 217 from:
`if(t) then`
to
`if(t and nor r) then`
This should cover Sticklord's concern, which is valid, I missed that one.

Rainrider:
To sum it up:
The concern about not being able to use .colorSmooth is valid, I'll address that.
The point for .colorPower is not. Except for the call to `UnitPowerType`, which just moved a scope up, it is all like it was.

elements/power.lua
ls-: Damn, I guess, I was way too tired, when I was editing first two comments yesterday.

We do need `t` beforehand, cuz we also need `t.atlas`, otherwise `.useAtlas` won't work in most cases T_T

Yeah, I stand corrected on this one 😒🔫

Sticklord:
Sorry to be a pain in the ass, but this don't address the issue when you do NOT specify any color option for the bar (colorClass, colorPower etc), that is when you do not want the Power element to apply a color to the bar.

Rainrider:
@Sticklord it's ok. I'll figure something out as we need to prevent coloring of .bg in that case too. Introducing a `.noColor` option would make things easier here I suppose, but this would require a change in layouts, which we probably want to avoid.

ls-:
Hey, that's me again, however, this time I'm not tired at all 😆

[Here's my solution](https://gist.github.com/ls-/ff0e1b2568b44f9414cdf03eb9e744c3).

Same test setup as before.
```Lua
local power = CreateFrame("StatusBar", nil, self)
power:SetAllPoints()
power:SetStatusBarTexture("Interface\\BUTTONS\\WHITE8X8")
power.colorPower = true
power.frequentUpdates = true
power.useAtlas = true

local bg = power:CreateTexture(nil, 'BACKGROUND')
bg:SetAllPoints(power)
bg:SetColorTexture(1, 1, 1, .4)
bg.multiplier = 0.2
power.bg = bg

self.Power = power
```

With `.powerColor = true`:
image

With no options at all:
image

I swapped few blocks around. This way widget resolves colour first, and then it decides between using atlas or normal texture. Colour's applied, if there's one.

Rainrider:
@ls- set `.colorSmooth` and `.useAtlas` to true and see what happens with the .bg

ls-:
@Rainrider,
pp_test

Rainrider:
This depends on `self.colors.smooth` then. My point is that this way the background colors differs from the  atlas.

ls-:
@Rainrider, but this way it respects settings, that's what @Sticklord asked for.

Anyway, we can check for `b` in atlas-block [here](https://gist.github.com/ls-/ff0e1b2568b44f9414cdf03eb9e744c3#file-test-lua-L49), and override `r, g, b`, if `b` isn't `nil`:
```Lua
if(b) then
	r, g, b = t[1], t[2], t[3]
end
```
otherwise we let it be as it's, so it won't be coloured.

ls-:
Here's my result:
image

No options and `.colorSmooth` respectively.

Rainrider:
For tapped or disconnected we probably don't want to overwrite the bg color for atlas.

ls-:
@Rainrider, well, you can remove [tapped/disconnected part](https://gist.github.com/ls-/ff0e1b2568b44f9414cdf03eb9e744c3#file-test-lua-L48), if you don't like it. I added it, cuz, IMHO, coloured bg and desaturated atlas look quite meh together...

As for "doing colour stuff for nothing". I don't see any problem there, cuz it's not for nothing, since A) atlas might not be there, and we'll use `r, g, b` to colour power bar's texture; B) it lets us know that some options are active, so we need to overwrite colours; C) if we decide not to do B part, it'll be used to colour bg.

ls-:
Here's diff between power.lua in master and mine:
```diff
diff --git a/elements/power.lua b/elements/power.lua
index c304d2b..79e66cb 100644
--- a/elements/power.lua
+++ b/elements/power.lua
@@ -108,7 +108,7 @@ for power, color in next, PowerBarColor do
 				oUF.colors.power[power][index] = {color.r, color.g, color.b}
 			end
 		else
-			oUF.colors.power[power] = {color.r, color.g, color.b}
+			oUF.colors.power[power] = {color.r, color.g, color.b, atlas = color.atlas}
 		end
 	end
 end
@@ -150,6 +150,7 @@ local Update = function(self, event, unit)
 	end
 	local cur, max = UnitPower(unit, displayType), UnitPowerMax(unit, displayType)
 	local disconnected = not UnitIsConnected(unit)
+	local tapped = not UnitPlayerControlled(unit) and UnitIsTapDenied(unit)
 	power:SetMinMaxValues(min or 0, max)
 
 	if(disconnected) then
@@ -159,7 +160,9 @@ local Update = function(self, event, unit)
 	end
 
 	power.disconnected = disconnected
+	power.tapped = tapped
 
+	local ptype, ptoken, altR, altG, altB = UnitPowerType(unit)
 	local r, g, b, t
 
 	if(power.colorTapping and not UnitPlayerControlled(unit) and UnitIsTapDenied(unit)) then
@@ -169,8 +172,6 @@ local Update = function(self, event, unit)
 	elseif(displayType == ALTERNATE_POWER_INDEX and power.altPowerColor) then
 		t = power.altPowerColor
 	elseif(power.colorPower) then
-		local ptype, ptoken, altR, altG, altB = UnitPowerType(unit)
-
 		t = self.colors.power[ptoken]
 		if(not t) then
 			if(power.GetAlternativeColor) then
@@ -194,7 +195,7 @@ local Update = function(self, event, unit)
 	elseif(power.colorReaction and UnitReaction(unit, 'player')) then
 		t = self.colors.reaction[UnitReaction(unit, "player")]
 	elseif(power.colorSmooth) then
-        local adjust = 0 - (min or 0)
+		local adjust = 0 - (min or 0)
 		r, g, b = self.ColorGradient(cur + adjust, max + adjust, unpack(power.smoothGradient or self.colors.smooth))
 	end
 
@@ -202,18 +203,33 @@ local Update = function(self, event, unit)
 		r, g, b = t[1], t[2], t[3]
 	end
 
-	if(b) then
-		power:SetStatusBarColor(r, g, b)
-
-		local bg = power.bg
-		if(bg) then
-			local mu = bg.multiplier or 1
-			bg:SetVertexColor(r * mu, g * mu, b * mu)
+	t = self.colors.power[ptoken or ptype]
+	local atlas = power.atlas or t.atlas
+	if(power.useAtlas and atlas and displayType ~= ALTERNATE_POWER_INDEX) then
+		power:SetStatusBarAtlas(atlas)
+		power:SetStatusBarColor(1, 1, 1)
+		if(power.colorTapping or power.colorDisconnected) then
+			power:GetStatusBarTexture():SetDesaturated(disconnected or tapped)
 		end
+		-- overwrite r, g, b, if there's some other colour
+		if(b) then
+			r, g, b = t[1], t[2], t[3]
+		end
+	else
+		power:SetStatusBarTexture(power.texture)
+		if(b) then
+			power:SetStatusBarColor(r, g, b)
+		end
+	end
+
+	local bg = power.bg
+	if(bg and b) then
+		local mu = bg.multiplier or 1
+		bg:SetVertexColor(r * mu, g * mu, b * mu)
 	end
 
 	if(power.PostUpdate) then
-		return power:PostUpdate(unit, cur, max, min)
+		return power:PostUpdate(unit, cur, max, min, ptoken, ptype)
 	end
 end
 
@@ -246,8 +262,9 @@ local Enable = function(self, unit)
 		-- For tapping.
 		self:RegisterEvent('UNIT_FACTION', Path)
 
-		if(power:IsObjectType'StatusBar' and not power:GetStatusBarTexture()) then
-			power:SetStatusBarTexture[[Interface\TargetingFrame\UI-StatusBar]]
+		if(power:IsObjectType'StatusBar') then
+			power.texture = power:GetStatusBarTexture() and power:GetStatusBarTexture():GetTexture() or [[Interface\TargetingFrame\UI-StatusBar]]
+			power:SetStatusBarTexture(power.texture)
 		end
 
 		return true
```

Rainrider:
Sorry Val, I missed line 48 in your code somehow. It's fine with me then, I don't have a better solution.

Would you mind pushing your changes to the branch here, I can squash and merge then if @p3lim has no objections.

ls-:
I hope I didn't miss anything 😒🔫

Rainrider:
Squashed and force pushed for a clean history. Pending approval :)

ls-:
@Rainrider, you'll have to squash one more time, forgot to fix one thingy T_T

p3lim:
Squash and merge at your own disposal, all looks good with the last few commits.

Rainrider:
Apparently the trailing white space are needed for haste's documentation tool

ls-:
TBH, at this point they're way too inconsistent... Many files don't have them anymore.

Issue #247, opened 2015-03-24
Power Color tag for Fontstrings


Kygos:
A tag similar to [raidcolor] but for power color (eg if Warrior then set FontColor to RGB 255, 0, 0)

Will help out rookies such as myself that can't grasp the PostUpdate Function quite yet.

p3lim:
You can easily add this tag yourself:

```Lua
oUF.Tags.Events['powercolor'] = 'UNIT_DISPLAYPOWER'
oUF.Tags.Methods['powercolor'] = function(unit)
	local _, powerType = UnitPowerType(unit)
	return Hex(_COLORS.power[powerType] or _COLORS.power.MANA)
end
```

PR #305, opened 2016-09-23
tags: Added 'powercolor' tag.

ls-:
Addresses request https://github.com/oUF-wow/oUF/issues/247.

### DIFF
--- a/elements/tags.lua
+++ b/elements/tags.lua
@@ -311,6 +311,25 @@ local tagStrings = {
 			return 'Affix'
 		end
 	end]],
+
+	['powercolor'] = [[function(u)
+		local pType, pToken, altR, altG, altB = UnitPowerType(u)
+		local t = _COLORS.power[pToken]
+
+		if(not t) then
+			if(altR) then
+				if(altR > 1 or altG > 1 or altB > 1) then
+					return Hex(altR / 255, altG / 255, altB / 255)
+				else
+					return Hex(altR, altG, altB)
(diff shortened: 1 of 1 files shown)


elements/tags.lua
```diff
@@ -311,6 +311,25 @@ local tagStrings = {
...
 	end]],
+
+	['powercolor'] = [[function(u)
+		local pType, pToken, altR, altG, altB = UnitPowerType(u)
+		local t = _COLORS.power[pToken]
+
+		if(not t) then
+			if(altR) then
+				if(altR > 1 or altG > 1 or altB > 1) then
+					t = {altR / 255, altG / 255, altB / 255}
```
p3lim: Return the value directly here, no need to create additional tables.
```Lua
return Hex(altR / 255, altG / 255, altB / 255)
```

elements/tags.lua
```diff
@@ -311,6 +311,25 @@ local tagStrings = {
...
+				else
+					t = {altR, altG, altB}
```
p3lim: Same as the comment above.

ls-:
Logic failed me, yet again... 😕

Rainrider:
You should also take power.GetAlternativeColor into account, so that the user could match the bar colours if they use it.

ls-:
@Rainrider, `GetAlternativeColor` is power bar's method defined in layout. You can't access such things from inside tag function O_o

Issue #297, opened 2016-09-12
ClassIcons: error when starting quest "Huln's War - The Arrival"


Blazeflack:
In the "Huln's War - The Arrival" quest you speak to an NPC which then transfers you to Azshara and force you into a vehicle state.

This causes the following error when playing on a Shadow Priest:
```
Message: ...e\AddOns\ElvUI\Libraries\oUF\elements\classicons.lua line 109:
   attempt to index field '?' (a nil value)
Debug:
   ...e\AddOns\ElvUI\Libraries\oUF\elements\classicons.lua:109:
      ...e\AddOns\ElvUI\Libraries\oUF\elements\classicons.lua:73
   (tail call): ?
   ...e\AddOns\ElvUI\Libraries\oUF\elements\classicons.lua:170:
      ...e\AddOns\ElvUI\Libraries\oUF\elements\classicons.lua:147
   (tail call): ?
   ElvUI\Libraries\oUF\ouf.lua:174: UpdateAllElements()
   ...ace\AddOns\ElvUI\Modules\unitframes\units\player.lua:193:
      ...ace\AddOns\ElvUI\Modules\unitframes\units\player.lua:80
   [C]: ?()
   ...rface\AddOns\ElvUI\Modules\unitframes\unitframes.lua:905: Update()
   ...rface\AddOns\ElvUI\Modules\unitframes\unitframes.lua:454: Update_AllFrames()
   ...rface\AddOns\ElvUI\Modules\unitframes\unitframes.lua:1125: ?()
   ...ibraries\CallbackHandler-1.0\CallbackHandler-1.0.lua:145:
      ...ibraries\CallbackHandler-1.0\CallbackHandler-1.0.lua:145
   [string "safecall Dispatcher[1]"]:4:
      [string "safecall Dispatcher[1]"]:4
   [C]: ?
   [string "safecall Dispatcher[1]"]:13: ?()
   ...ibraries\CallbackHandler-1.0\CallbackHandler-1.0.lua:90: Fire()
   ...AddOns\ElvUI\Libraries\AceEvent-3.0\AceEvent-3.0.lua:120:
      ...AddOns\ElvUI\Libraries\AceEvent-3.0\AceEvent-3.0.lua:119
Locals:
None
```

I added some debug prints to try to understand what is happening, and this is the result:
```
VisibilityPath: PLAYER_ENTERING_WORLD player
Path: ClassPowerEnable vehicle COMBO_POINTS
[...]

PR #307, opened 2016-09-26
core: Make sure UpdateAllElements has an event

p3lim:
Not providing events when calling UAE from 3rd-party sources
makes it very hard to debug should something go wrong.

Relevant issue: https://github.com/oUF-wow/oUF/issues/297#issuecomment-246696419.

### DIFF
--- a/ouf.lua
+++ b/ouf.lua
@@ -145,7 +145,7 @@ for k, v in pairs{
 		local unit = self.unit
 		if(not UnitExists(unit)) then return end
 
-		assert(type(event) == 'string', 'Missing argument "event" in UpdateAllElements.')
+		assert(type(event) == 'string', 'Invalid argument "event" in UpdateAllElements.')
 
 		if(self.PreUpdate) then
 			self:PreUpdate(event)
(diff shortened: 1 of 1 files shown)


ls-:
TBH, I'd use assert o_O
```Lua
assert(event, 'Missing argument "event" in UpdateAllElements.')
```

p3lim:
You're right.

ls-:
Hm, one more thing. To make things easier to debug, I suggest adding frame's name to error message.
```Lua
assert(type(event) == 'string', 'Missing argument "event" in '..self:GetDebugName()..':UpdateAllElements().')
```
Sorry 😢🔫

p3lim:
Here's the stack without that addition:
```
1x oUF\ouf-1.6.9.lua:148: Missing argument "event" in UpdateAllElements.
[C]: in function `assert'
oUF\ouf-1.6.9.lua:148: in function `UpdateAllElements'
[string "oUF_P3limPlayer:UpdateAllElements()"]:1: in main chunk
```

ls-:
Nvm, I suggest retarded things sometimes...

PR #310, opened 2016-09-29
runebar: Set cooldown start time to 0 if rune was energised

ls-:
When CD ends, start time is set to 0, however it doesn't happen, when rune is energised, so we pass wrong info to layout, cuz we don't differentiate between these two events.

Proof:
image

### DIFF
--- a/elements/runebar.lua
+++ b/elements/runebar.lua
@@ -83,7 +83,7 @@ local Update = function(self, event, rid, energized)
 	end
 
 	if(runes.PostUpdate) then
-		return runes:PostUpdate(rune, rid, start, duration, energized or runeReady)
+		return runes:PostUpdate(rune, rid, energized and 0 or start, duration, energized or runeReady)
 	end
 end
 


Rainrider:
Out of curiosity what use case was broken by that?

ls-:
In my UI when rune is recharged I [animate and recolour it](https://github.com/ls-/ls_UI/blob/master/modules/unitframes/elements/classpower.lua#L408-L413), but sometimes it wasn't happening, so few days ago I decided to investigate this issue.

My a bit outdated `runeReady and start == 0` check was causing it, `start == 0` part is there since pre-Legion times, and probably should have been removed by now, but it's needed in the past to avoid false triggering.

I decided to add this change to make things a bit more consistent on our end anyway. So this one was 75% cosmetic and 25% preventive update 😓

I asked p3lim if it's PRworthy, he asked for proof it's actually happening, that's why I included that screenshot.

Sorry for littering git log 😞

Rainrider:
Strange, I've been using that for 3 years now and never noticed a problem https://github.com/Rainrider/oUF_Rain/commit/dcfe75fad3e9f708282c87ff730844888f3ef278

ls-:
TBF, it's quite easy not to notice it.

PR #313, opened 2016-10-10
castbar: add a .holdTime option

Rainrider:
Also use `SetColorTexture` instead of `SetTexture` for setting the color.

This is the minimal approach so that nothing changes for current implementations layout wise,

However it is not optimal. The way to use it is to set `Castbar.holdTime` to `GetTime() + CONSTANT` in both `Castbar.PostCastFailed` and `Castbar.PostCastInterrupted` and show the castbar again. The user can also update the castbar text and max value if desired. Also, with the current behavior the bar will be hidden first (UNIT_SPELLCAST_STOP fires after the first UNIT_SPELLCAST_INTTERUPTED, which fires 4 times).

A more intuitive approach would be to handle this from the default element (including setting the castbar text to `_G.INTERRUPTED`/`_G.FAILED` to mimic the default UI) and leave the castbar hiding to the OnUpdate script.

What do you think?

PS: `Castbar.interrupt` actually meaning `Castbar.notInterruptible` drives me crazy :)

### DIFF
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -111,7 +111,7 @@ local UNIT_SPELLCAST_START = function(self, event, unit, spell)
 	if(self.unit ~= unit and self.realUnit ~= unit) then return end
 
 	local castbar = self.Castbar
-	local name, _, text, texture, startTime, endTime, _, castid, interrupt = UnitCastingInfo(unit)
+	local name, _, text, texture, startTime, endTime, _, castid, notInterruptible = UnitCastingInfo(unit)
 	if(not name) then
 		castbar:Hide()
 		return
@@ -126,7 +126,8 @@ local UNIT_SPELLCAST_START = function(self, event, unit, spell)
 	castbar.max = max
 	castbar.delay = 0
 	castbar.casting = true
-	castbar.interrupt = interrupt
+	castbar.interrupt = notInterruptible -- NOTE: deprecated; to be removed
+	castbar.notInterruptible = notInterruptible
 	castbar.holdTime = 0
 
 	castbar:SetMinMaxValues(0, max)
@@ -137,7 +138,7 @@ local UNIT_SPELLCAST_START = function(self, event, unit, spell)
 	if(castbar.Time) then castbar.Time:SetText() end
 
 	local shield = castbar.Shield
-	if(shield and interrupt) then
+	if(shield and notInterruptible) then
 		shield:Show()
 	elseif(shield) then
 		shield:Hide()
@@ -167,7 +168,8 @@ local UNIT_SPELLCAST_FAILED = function(self, event, unit, spellname, _, castid)
 	end
 
 	castbar.casting = nil
-	castbar.interrupt = nil
+	castbar.interrupt = nil -- NOTE: deprecated; to be removed
+	castbar.notInterruptible = nil
 	castbar:SetValue(0)
 	castbar:Hide()
 
@@ -251,7 +253,8 @@ local UNIT_SPELLCAST_STOP = function(self, event, unit, spellname, _, castid)
 	end
 
 	castbar.casting = nil
-	castbar.interrupt = nil
+	castbar.interrupt = nil -- NOTE: deprecated; to be removed
+	castbar.notInterruptible = nil
 	castbar:SetValue(0)
 	castbar:Hide()
 
@@ -264,7 +267,7 @@ local UNIT_SPELLCAST_CHANNEL_START = function(self, event, unit, spellname)
 	if(self.unit ~= unit and self.realUnit ~= unit) then return end
 
 	local castbar = self.Castbar
-	local name, _, text, texture, startTime, endTime, isTrade, interrupt = UnitChannelInfo(unit)
+	local name, _, text, texture, startTime, endTime, isTrade, notInterruptible = UnitChannelInfo(unit)
 	if(not name) then
 		return
 	end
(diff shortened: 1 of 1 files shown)


Rainrider:
Here the PostCast* functions I use for reference:

```lua
local PostCastFailed = function(castbar, unit, name, castID)
	castbar:SetStatusBarColor(0.69, 0.31, 0.31)
	castbar:SetValue(castbar.max)
	local text = castbar.Text
	if text then
		text:SetText(_G.FAILED)
	end
	local time = castbar.Time
	if time then
		time:SetText(name)
	end
	castbar.holdTime = GetTime() + 1
	castbar:Show()
end

local PostCastInterrupted = function(castbar, unit, name, castID)
	castbar:SetStatusBarColor(0.69, 0.31, 0.31)
	castbar:SetValue(castbar.max)
	local text = castbar.Text
	if text then
		text:SetText(_G.INTERRUPTED)
	end
	local time = castbar.Time
	if time then
		time:SetText(name)
	end
	castbar:Show()
	castbar.holdTime = GetTime() + 1
end
```

elements/castbar.lua
```diff
@@ -479,7 +490,7 @@ local Enable = function(object, unit)
 
 		local sz = castbar.SafeZone
 		if(sz and sz:IsObjectType'Texture' and not sz:GetTexture()) then
-			sz:SetTexture(1, 0, 0)
+			sz:SetColorTexture(1, 0, 0)
```
p3lim: Split into it's own comment, it's irrelevant to the feature

p3lim:
> PS: Castbar.interrupt actually meaning Castbar.notInterruptible drives me crazy :)

Add it (without replacing the original for compat) and we'll deprecate the original in next next.

Rainrider:
Took the liberty to clean up a bit too.

elements/castbar.lua
```diff
@@ -183,6 +188,11 @@ local UNIT_SPELLCAST_INTERRUPTED = function(self, event, unit, spellname, _, cas
 		return
 	end
 
+	local text = castbar.Text
+	if(text) then
+		text:SetText(INTERRUPTED)
+	end
+
 	castbar.casting = nil
```
p3lim: Sure you don't want to leave these two to the layout?

elements/castbar.lua
```diff
@@ -146,56 +153,66 @@ local UNIT_SPELLCAST_START = function(self, event, unit, spell)
...
 	local castbar = self.Castbar
 	if(castbar.castid ~= castid) then
 		return
 	end
 
+	local text = castbar.Text
+	if(text) then
+		text:SetText(FAILED)
+	end
+
```
p3lim: Sure you don't want to leave this to the layout?

elements/castbar.lua
Rainrider: The user probably won't even notice that if they don't use the .holdTime option because the bar will be hidden by the OnUpdate script on the next frame. The change allows to use the same PostUpdate hook for UNIT_SPELLCAST_FAILED and _INTERRUPTED.

I would even go further and set .holdTime to `GetTime() + a layout defined constant` so that it would be possible to eliminate the necessity for a PostUpdate hook (unless they want coloring or something)

I'm not sure about setting the bar value to its max though. While this is the default ui behavior, maybe it is interesting to see when the cast was interrupted.

elements/castbar.lua
Rainrider: See https://github.com/oUF-wow/oUF/pull/313#discussion_r82776603

elements/castbar.lua
p3lim: Fair enough, setting the bar to max can be done in PostUpdate should the layout _really_ want that for the looks (the other way around is harder).

elements/castbar.lua
```diff
@@ -120,7 +125,9 @@ local UNIT_SPELLCAST_START = function(self, event, unit, spell)
 	castbar.max = max
 	castbar.delay = 0
 	castbar.casting = true
-	castbar.interrupt = interrupt
+	castbar.interrupt = notInterruptible -- NOTE: deprecated; to be removed
+	castbar.notInterruptible = notInterruptible
+	castbar.holdTime = 0
```
p3lim: This will reset the option set by the layout.

elements/castbar.lua
```diff
@@ -431,29 +450,30 @@ local ForceUpdate = function(element)
...
+			self:RegisterEvent("UNIT_SPELLCAST_INTERRUPTED", UNIT_SPELLCAST_INTERRUPTED)
+			self:RegisterEvent("UNIT_SPELLCAST_INTERRUPTIBLE", UNIT_SPELLCAST_INTERRUPTIBLE)
+			self:RegisterEvent("UNIT_SPELLCAST_NOT_INTERRUPTIBLE", UNIT_SPELLCAST_NOT_INTERRUPTIBLE)
+			self:RegisterEvent("UNIT_SPELLCAST_DELAYED", UNIT_SPELLCAST_DELAYED)
+			self:RegisterEvent("UNIT_SPELLCAST_CHANNEL_START", UNIT_SPELLCAST_CHANNEL_START)
+			self:RegisterEvent("UNIT_SPELLCAST_CHANNEL_UPDATE", UNIT_SPELLCAST_CHANNEL_UPDATE)
+			self:RegisterEvent("UNIT_SPELLCAST_CHANNEL_STOP", UNIT_SPELLCAST_CHANNEL_STOP)
 		end
 
+		castbar.holdTime = 0
```
p3lim: This will also reset the option set by the layout.

elements/castbar.lua
Rainrider: This is intended as it is the default ui behavior.

elements/castbar.lua
Rainrider: This is intended as it is the default ui behavior.

Rainrider:
Latest change removes the necessity for the layout to call `GetTime()` to set .holdTime properly. It would also eliminate the need of PostUpdate hook to _INTERRUPT and _FAILED unless they want special handling. It also keeps the opt-in behavior so no change in current layouts is required for the element to work.

p3lim:
Do **not** squash merge this PR.

PR #318, opened 2016-10-17
GetCreatureDifficultyColor and 199

zorker:
QuestDifficultyColor does not scale with timewalking dungeons.
99 is outdated for the unknown max level. Use 199 instead.


Talyrius:
It should be 999 instead of 199, IMO.

p3lim:
Yeah I modified it: https://github.com/oUF-wow/oUF/commit/24c8f0660d1a65efe82aa9d5c2293f3eb1a458fa

PR #329, opened 2017-01-22
core: Expose header visibility

p3lim:
Unless anyone knows a way of getting this without exposing it here, SecureStateDriver is pretty air-tight.


haste:
There's no other way of reliably getting the visibility state of a frame from the SecureStateDrive, but why do you need to expose this?

p3lim:
Making a plugin that can enable/disable objects and headers based on spec role.

haste:
We would still have to save the visibility somewhere then.

PR #331, opened 2017-02-23
pvpindicator: Swap factionGroup only for player

ls-:
TSIA.

### DIFF
--- a/elements/pvpindicator.lua
+++ b/elements/pvpindicator.lua
@@ -58,7 +58,7 @@ local function Update(self, event, unit)
 	local factionGroup = UnitFactionGroup(unit)
 
 	if(UnitIsPVPFreeForAll(unit)) then
-		if(prestigeLevel > 0 and element.Prestige) then
+		if(element.Prestige and prestigeLevel > 0) then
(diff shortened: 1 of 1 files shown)


elements/pvpindicator.lua
```diff
@@ -71,7 +71,7 @@ local function Update(self, event, unit)
 
 		status = 'ffa'
 	elseif(factionGroup and factionGroup ~= 'Neutral' and UnitIsPVP(unit)) then
-		if(UnitIsMercenary(unit)) then
+		if(unit == 'player' and UnitIsMercenary(unit)) then
```
Rainrider: Why this?

ls-: Ask Blizzard T_T In merc mode `UnitFactionGroup` returns real faction group for player unit, but it works fine for other units, so there's no need to swap anything for them.

ls-: For more info, see [here](https://github.com/Gethe/wow-ui-source/blob/live/FrameXML/PlayerFrame.lua#L150-L157), this swap is unique to player frame.

PR #334, opened 2017-02-27
stagger: Revamp Update function.

ls-:
- Made it more fail-safe;
- Removed unnecessary var caching;
- Reordered PostUpdate arg list. Throughout oUF we always use `cur, max, ...` arg order.

### DIFF
--- a/elements/stagger.lua
+++ b/elements/stagger.lua
@@ -30,10 +30,11 @@ The default StatusBar texture will be applied if the UI widget doesn't have a st
     self.Stagger = Stagger
 --]]
 
+if select(2, UnitClass('player')) ~= 'MONK' then return end
+
 local parent, ns = ...
 local oUF = ns.oUF
 
-local _, playerClass = UnitClass("player")
 
 local function Update(self, event, unit)
 	if(unit and unit ~= self.unit) then return end
@@ -139,8 +140,6 @@ local function ForceUpdate(element)
 end
 
 local function Enable(self, unit)
-	if(playerClass ~= 'MONK') then return end
-
 	local element = self.Stagger
 	if(element) then
 		element.__owner = self
--- a/elements/stagger.lua
+++ b/elements/stagger.lua
@@ -50,8 +50,8 @@ local function Update(self, event, unit)
 		element:PreUpdate()
 	end
 
-	local cur = UnitStagger("player")
-	local max = UnitHealthMax("player")
+	local cur = UnitStagger('player')
+	local max = UnitHealthMax('player')
 	local perc = cur / max
 
 	element:SetMinMaxValues(0, max)
--- a/elements/stagger.lua
+++ b/elements/stagger.lua
@@ -33,19 +33,7 @@ The default StatusBar texture will be applied if the UI widget doesn't have a st
 local parent, ns = ...
 local oUF = ns.oUF
 
--- percentages at which the bar should change color
-local STAGGER_YELLOW_TRANSITION = STAGGER_YELLOW_TRANSITION
-local STAGGER_RED_TRANSITION = STAGGER_RED_TRANSITION
-
--- table indices of bar colors
-local STAGGER_GREEN_INDEX = STAGGER_GREEN_INDEX or 1
-local STAGGER_YELLOW_INDEX = STAGGER_YELLOW_INDEX or 2
-local STAGGER_RED_INDEX = STAGGER_RED_INDEX or 3
-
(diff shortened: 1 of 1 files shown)


elements/stagger.lua
```diff
@@ -33,19 +33,7 @@ The default StatusBar texture will be applied if the UI widget doesn't have a st
...
 local oUF = ns.oUF
 
--- percentages at which the bar should change color
-local STAGGER_YELLOW_TRANSITION = STAGGER_YELLOW_TRANSITION
-local STAGGER_RED_TRANSITION = STAGGER_RED_TRANSITION
-
--- table indices of bar colors
-local STAGGER_GREEN_INDEX = STAGGER_GREEN_INDEX or 1
-local STAGGER_YELLOW_INDEX = STAGGER_YELLOW_INDEX or 2
-local STAGGER_RED_INDEX = STAGGER_RED_INDEX or 3
```
haste: We usually keep these around incase Blizzard decides to remove them. It's happened a couple of times before, so better safe than sorry.

elements/stagger.lua
```diff
@@ -61,45 +49,52 @@ local function Update(self, event, unit)
...
+		r, g, b = t[1], t[2], t[3]
+	end
+
+	if(b) then
+		element:SetStatusBarColor(r, g, b)
+	end
 
 	local bg = element.bg
-	if(bg) then
+	if(bg and b) then
```
haste: Might as well move the bg check up into the if above.

elements/stagger.lua
```diff
@@ -61,45 +49,52 @@ local function Update(self, event, unit)
...
+	* cur            - the amount of staggered damage
+	* max            - the player's maximum possible health value
+	* perc           - the amount of staggered damage relative to the player's maximum health
+	* r              - the red component of the StatusBar color (depends on perc)
+	* g              - the green component of the StatusBar color (depends on perc)
+	* b              - the blue component of the StatusBar color (depends on perc)
 	--]]
 	if(element.PostUpdate) then
-		element:PostUpdate(maxHealth, stagger, staggerPercent, r, g, b)
+		element:PostUpdate(cur, max, perc, r, g, b)
```
haste: perc is pretty useless here, might as well remove it while we're cleaning stuff up

elements/stagger.lua
ls-: K, I'll read these tomorrow, mb I should (re)add these vars too?
```Lua
local BREWMASTER_POWER_BAR_NAME = BREWMASTER_POWER_BAR_NAME or "STAGGER"
local STAGGER_YELLOW_TRANSITION =  STAGGER_YELLOW_TRANSITION or 0.3
local STAGGER_RED_TRANSITION = STAGGER_RED_TRANSITION or 0.6
```

elements/stagger.lua
ls-: Well, makes sense. Actually, I'll consolidate all three ifs :p 
```Lua
local r, g, b
if(t) then
	r, g, b = t[1], t[2], t[3]
	if(b) then
		element:SetStatusBarColor(r, g, b)

		local bg = element.bg
		if(bg and b) then
			local mu = bg.multiplier or 1
			bg:SetVertexColor(r * mu, g * mu, b * mu)
		end
	end
end
```

elements/stagger.lua
ls-: TBF, I'm not even sure if we need/want to pass `r, g, b`, we could pass stagger index instead or nothing at all :D

haste: Passing r, g ,b is fine. It's much more of a hassle to get those values than... `cur / max`.

elements/stagger.lua
p3lim: To be fair, if Blizzard decides to remove them, the element probably will need to be changed anyways.

haste: They've moved stuffed away from "consts" before, without changing the value. It's a tiny safe guard if that should happen again. If it breaks it breaks, but no need to increase the odds when we can mitigate them without much hassle.

elements/stagger.lua
Rainrider: Agree on that

elements/stagger.lua
Rainrider: Those are actually more for readability then anything else.

ls-:
It's possible to auto-squash them upon merge, that's why I didn't add `stagger:` prefix to later commits.

Issue #108, opened 2011-12-05
AltPowerBar Buff Timers


Haleth:
Seems like a new interface element introduced in 4.3. It appears at some Darkmoon Faire games and shares a lot with the alt power bar, but both can be shown at the same time.

https://github.com/tekkub/wow-ui-source/blob/ptr/FrameXML/UnitPowerBarAlt.lua

http://img847.imageshack.us/img847/5107/wowscrnshot120511095336.jpg

Rainrider:
Is the Darkmoon Fair the only thing that uses the couter bar system?

Rainrider:
Please take a look at https://github.com/oUF-wow/oUF/tree/player-buff-timer for a WIP implementation of this. I opted for a separate element and pushed it unfinished in order to get feedback before the Darkmoon Fair ends.

To be yet included:
  - [ ] tooltips?
  - [ ] documentation
  - [ ] PostCreateTimer callback?
  - [ ] timer placement options (see TODOs in the code)

I only found the 3 darkmoon fair quests ("It's Hammer Time", "He Shoots, He Scores!" and "Tonk Commander") that use this. And only "Tonk Commander" does not have a normal buff to track the remaining time, so it's actually only one quest that "really" needs this.

Minimal layout implementation is simply `self.PlayerBuffTimers = {}`, where self is the player frame.

You could also use something like that if you prefer normal cooldown buttons over bars:
```lua
self.PlayerBuffTimers.CreateTimer = function(self, duration, expiration, auraID)
	local name, _, texture = GetSpellInfo(auraID)
	local button = CreateFrame("Button", nil, self)
	button:SetSize(32, 32)
	local cd = CreateFrame("Cooldown", nil, button, "CooldownFrameTemplate")
	cd:SetAllPoints()
	cd:SetReverse(true)
	local icon = button:CreateTexture(nil, "BORDER")
	icon:SetAllPoints()
	icon:SetTexture(texture)
	cd:SetCooldown(expiration - duration, duration)
	
	return button
end
```

PR #336, opened 2017-03-06
Player buff timers

Rainrider:
See #108 

Pending:
- [x] display current/min/max values
- [x] documentation

If you want to test outside of Darkmoon, uncomment UnitPowerBarTimerInfo (above the Update function) and `SetPosition(timer:GetParent())` in `OnUpdate` (if you care about positioning) and use the elements `:ForceUpdate`.

### DIFF
--- a/elements/playerbufftimers.lua
+++ b/elements/playerbufftimers.lua
@@ -15,11 +15,8 @@ local function CreateTimer(self, duration, expiration, auraID)
 	local element = self.PlayerBuffTimers
 	local timer = CreateFrame('StatusBar', nil, self)
 	timer:SetMinMaxValues(0, duration)
-	-- TODO: let the layout deside?
-	timer:SetStatusBarTexture([[Interface\TargetingFrame\UI-StatusBar]])
-	timer:SetHeight(10)
-	timer:SetPoint('LEFT')
-	timer:SetPoint('RIGHT')
+	timer:SetStatusBarTexture([[Interface\TargetingFrame\UI-StatusBar]]) -- TODO: let the layout deside
+	timer:SetSize(element.width or element:GetWidth(), element.height or 10)
 	timer:SetScript('OnUpdate', OnUpdate)
 
 	if(element.PostCreateTimer) then
@@ -41,6 +38,40 @@ local function GetTimer(self, duration, expiration, auraID)
 	return (self.PlayerBuffTimers.CreateTimer or CreateTimer)(self, duration, expiration, auraID)
 end
 
+local growthDirection = {
+	TOPLEFT     = { 1, -1},
+	TOPRIGHT    = {-1, -1},
+	BOTTOMLEFT  = { 1,  1},
+	BOTTOMRIGHT = {-1,  1},
+}
+
+local function SetPosition(element)
+	local timers = element.timers
+
+	local width = (element.width or element:GetWidth()) + (element['spacing-x'] or element.spacing or 0)
+	local height = (element.height or 10) + (element['spacing-y'] or element.spacing or 0)
+	local anchor = element.anchor or 'TOPLEFT'
+	local direction = growthDirection[anchor] -- TODO: error on invalid anchor
+	local x = direction[1]
+	local y = direction[2]
+	local cols = math.floor(element:GetWidth() / width + .5)
+	local rows = math.floor(element:GetHeight() / height + .5)
+
+	for _, timer in next, timers do
+		local i = timer.index
+		local col, row
+		if(element.primaryAxis == 'x') then
+			col = (i - 1) % cols
+			row = math.floor((i - 1) / cols)
+		else
+			col = math.floor((i - 1) / rows)
+			row = (i - 1) % rows
+		end
+		timer:ClearAllPoints()
+		timer:SetPoint(anchor, element, anchor, col * width * x, row * height * y)
+	end
+end
+
 local function Update(self, event, unit)
 	local element = self.PlayerBuffTimers
 	local timers = element.timers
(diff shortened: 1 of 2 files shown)


elements/playerbufftimers.lua
```diff
@@ -0,0 +1,290 @@
...
+	end
+end
+
+local function Path(self, ...)
+	--[[ Override: PlayerBuffTimers:Override(...)
+	Used to completely override the internal update function.
+
+	* self - the PlayerBuffTimers widget
+	* ...  - the event and the arguments that accompany it
+	--]]
```
p3lim: See the documentation changes to overrides I did in the cleanup/review PR.

elements/playerbufftimers.lua
```diff
@@ -0,0 +1,290 @@
...
+
+local function UpdateTimer(timer, duration, expiration, barID, auraID)
+	timer:SetMinMaxValues(0, duration)
+	timer.duration = duration
+	timer.expiration = expiration
+	timer.auraID = auraID
+end
+
+local function CreateTimer(element, index)
+	local timer = CreateFrame('StatusBar', nil, element)
```
p3lim: Let the layout create the timers, and let them create how many they want, just like with ClassPower and Totems.
The layout should always be in control of frame creation.

elements/playerbufftimers.lua
```diff
@@ -0,0 +1,290 @@
...
+	timer:SetStatusBarTexture([[Interface\TargetingFrame\UI-StatusBar]]) -- TODO: let the layout deside
```
p3lim: The above comment solves this issue.

elements/playerbufftimers.lua
```diff
@@ -0,0 +1,290 @@
...
+	timer:SetSize(element.width or element:GetWidth(), element.height or 10)
+	timer:SetScript('OnUpdate', onUpdate)
+	timer:SetScript('OnEnter', onEnter)
+	timer:SetScript('OnLeave', onLeave)
```
p3lim: You can check if the layout set these with `:HasScript()`, then set yours if the layout didn't.

elements/playerbufftimers.lua
```diff
@@ -0,0 +1,290 @@
...
+		--SetPosition(timer:GetParent()) -- TODO: remove after testing
+	end
+end
+
+local function UpdateTimer(timer, duration, expiration, barID, auraID)
+	timer:SetMinMaxValues(0, duration)
+	timer.duration = duration
+	timer.expiration = expiration
+	timer.auraID = auraID
+end
```
p3lim: Merge this into the Update function instead.

p3lim:
I'm also not sure about the name, it may be misleading to layout creators.

p3lim:
Also make sure you kill `PlayerBuffTimerManager` (the default implementation) in Enable and return it with Disable.

elements/playerbufftimers.lua
Rainrider: This is not the case in the aura element. Auras are created and then re-used on the fly.

elements/playerbufftimers.lua
Rainrider: The layout has to provide `:UpdateTimer` if they use a widget type other than StatusBar. The code is like this to provide a reference implementation. I could move the attributes to the Update function though.

elements/playerbufftimers.lua
p3lim: Auras are in charge of a lot of extra widgets, this one only has one, and is very comparable to elements like ClassPower, Totems and Runes. Runes and this one is pretty much identical apart from the updating, yet it lets the layout create the bars.

elements/playerbufftimers.lua
Rainrider: Missed that one, you are right.

elements/playerbufftimers.lua
Rainrider: Will fix

elements/playerbufftimers.lua
```diff
@@ -0,0 +1,290 @@
...
+
+local function ForceUpdate(element)
+	return Path(element.__owner, 'ForceUpdate', 'player')
+end
+
+local function Enable(self, unit)
+	if(unit ~= 'player') then return end
+	local element = self.PlayerBuffTimers
+	if(not element) then return end
```
Rainrider: This is just me yes - avoid table access if you don't need it and save indents if you can. Can change for consistency, not a problem.

elements/playerbufftimers.lua
Rainrider: Yes, but the amount of max runes is known in advance.

p3lim: Which is where the similarities from ClassPower comes in, the layout creates as many as necessary for their support, not the layout.
While ClassPower has a theoretical max (at 10), this element will probably not see a number that high.

Rainrider: I see your point. I'd like to know what @ls- thinks on that too.

Rainrider:
> I'm also not sure about the name, it may be misleading to layout creators.
> Yes, I'm aware that blizzard uses this name too.
It is about timers for buffs on the player. By the looks of the default code, the system was meant to be expanded and thus the generic name. The default implementation however takes too much screen estate, which is maybe the reason they haven't done anything else with it yet.

elements/playerbufftimers.lua
ls-: TBH, I think these timers are quite similar to aura buttons, essentially they are aura indicator that look like status bars or whatever.

There's a way to override `CreateTimer` function, @Rainrider just has to make sure that if it's overridden, and something is missing, e.g. `UpdateTimer` method, it's not called/changed/etc.

Rainrider: `UpdateTimer` is mandatory as I see it, because the code in it depends on the used widget and how the user laid it out (`:SetValue` for StatusBar, `:SetCooldown` for Cooldown, `:SetText` or `:SetFormattedText` for FontStrings, etc).

Rainrider:
Addressed the requested changes in #356

PR #340, opened 2017-03-10
Cleanup/review

p3lim:
Misc cleanup for the 7.0 release, push anything you find to this branch.

### DIFF
--- a/elements/additionalpower.lua
+++ b/elements/additionalpower.lua
@@ -114,7 +114,7 @@ local function Path(self, ...)
 	--[[ Override: AdditionalPower:Override(event, unit, ...)
 	Used to completely override the element's update process.
 
-	* self  - the AdditionalPower element
+	* self  - the parent object
 	* event - the event triggering the update
 	* unit  - the unit accompanying the event
 	* ...   - the arguments accompanying the event
--- a/elements/additionalpower.lua
+++ b/elements/additionalpower.lua
@@ -111,7 +111,7 @@ local function Update(self, event, unit, powertype)
 end
 
 local function Path(self, ...)
-	--[[ Override: AdditionalPower:Override(event, unit, ...)
+	--[[ Override: AdditionalPower.Override(self, event, unit, ...)
 	Used to completely override the element's update process.
 
 	* self  - the parent object
--- a/elements/alternativepower.lua
+++ b/elements/alternativepower.lua
@@ -93,7 +93,7 @@ local function Path(self, ...)
 	--[[ Override: AlternativePower:Override(event, unit, ...)
 	Used to completely override the element's update process.
 
-	* self  - the AlternativePower element
+	* self  - the parent object
 	* event - the event triggering the update (string)
 	* unit  - the unit accompanying the event (string)
 	* ...   - the arguments accompanying the event
--- a/elements/alternativepower.lua
+++ b/elements/alternativepower.lua
@@ -90,7 +90,7 @@ local function Update(self, event, unit, powerType)
 end
 
 local function Path(self, ...)
-	--[[ Override: AlternativePower:Override(event, unit, ...)
+	--[[ Override: AlternativePower.Override(self, event, unit, ...)
 	Used to completely override the element's update process.
 
 	* self  - the parent object
--- a/elements/assistantindicator.lua
+++ b/elements/assistantindicator.lua
@@ -60,7 +60,7 @@ local function Path(self, ...)
 	--[[ Override: AssistantIndicator:Override(event, ...)
 	Used to completely override the element's update process.
 
-	* self  - the AssistantIndicator element
+	* self  - the parent object
 	* event - the event triggering the update (string)
 	* ...   - the arguments accompanying the event (string)
 	--]]
--- a/elements/assistantindicator.lua
+++ b/elements/assistantindicator.lua
@@ -57,7 +57,7 @@ local function Update(self, event)
 end
 
 local function Path(self, ...)
-	--[[ Override: AssistantIndicator:Override(event, ...)
+	--[[ Override: AssistantIndicator.Override(self, event, ...)
 	Used to completely override the element's update process.
 
 	* self  - the parent object
--- a/blizzard.lua
+++ b/blizzard.lua
@@ -1,6 +1,15 @@
 local parent, ns = ...
 local oUF = ns.oUF
 
+-- sourced from Blizzard_ArenaUI/Blizzard_ArenaUI.lua
+local MAX_ARENA_ENEMIES = MAX_ARENA_ENEMIES or 5
+
+-- sourced from FrameXML/TargetFrame.lua
+local MAX_BOSS_FRAMES = MAX_BOSS_FRAMES or 5
+
+-- sourced from FrameXML/PartyMemberFrame.lua
+local MAX_PARTY_MEMBERS = MAX_PARTY_MEMBERS or 4
+
 local hiddenParent = CreateFrame('Frame', nil, UIParent)
 hiddenParent:SetAllPoints()
 hiddenParent:Hide()
@@ -78,7 +87,7 @@ function oUF:DisableBlizzard(unit)
 		if(id) then
 			handleFrame('Boss' .. id .. 'TargetFrame')
 		else
-			for i = 1, 5 do
+			for i = 1, MAX_BOSS_FRAMES do
 				handleFrame(string.format('Boss%dTargetFrame', i))
 			end
 		end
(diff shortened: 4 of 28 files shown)


elements/additionalpower.lua
```diff
@@ -98,10 +100,10 @@ local function Update(self, event, unit, powertype)
 	--[[ Callback: AdditionalPower:PostUpdate(unit, cur, max)
 	Called after the element has been updated.
 
-	* self - the AdditionalPower element
-	* unit - the event unit that the update has been triggered for
-	* cur  - the current value of the player's additional power
-	* max  - the maximum possible value of the player's additional power
+	* self - AdditionalPower element
+	* unit - the unit for which the update has been triggered
```
Rainrider: ha! you forgot a `the` here :)

elements/additionalpower.lua
```diff
@@ -1,31 +1,32 @@
...
 # Element: Additional Power Bar
 
-Handles updating and visibility of a status bar displaying the player's additional power, such as Mana for Balance
+Handles updating and visibility of a status bar that displays player's additional power, such as Mana for Balance
 druids.
 
 ## Widget
 
-AdditionalPower - A StatusBar to represent the player's current additional power.
+AdditionalPower - A StatusBar that is used to represent player's additional power.
```
p3lim: "A StatusBar that is used to display the player's additional power."

elements/additionalpower.lua
```diff
@@ -1,31 +1,32 @@
...
 
 ## Sub-Widgets
 
-.bg - A Texture, which functions as a background. It will inherit the color of the main StatusBar.
+.bg - A Texture that is used as a background. Inherits the widget's color.
```
p3lim: element, not widget.

elements/additionalpower.lua
```diff
@@ -1,31 +1,32 @@
...
 
 ## Notes
 
-The default StatusBar texture will be applied if the UI widget doesn't have a status bar texture or color defined.
+Default status bar texture will be applied if the widget doesn't have a texture or a color set.
```
p3lim: "The default status bar texture will be applied if the element doesn't have a texture or a color set."

elements/additionalpower.lua
```diff
@@ -1,31 +1,32 @@
...
 
 ## Options
 
-.colorClass  - Use `self.colors.class[class]` to color the bar.
-.colorSmooth - Use `self.colors.smooth` to color the bar with a smooth gradient based on the players current additional
+.colorClass  - Uses `self.colors.class[class]` to color the bar based on player's class.
+.colorSmooth - Uses `self.colors.smooth` to color the bar with a smooth gradient based on player's current additional
```
p3lim: **The** player val.

p3lim: And I don't agree with the change from "use" to "uses" on any of these.

elements/additionalpower.lua
ls-: "A StatusBar that is used to display the player's additional power." would be better imho...

elements/additionalpower.lua
ls-: Hm... Why? O_o Widget means status bar.

elements/alternativepower.lua
```diff
@@ -159,7 +158,7 @@ local function Enable(self, unit)
 			end
 
 			--[[ Override: AlternativePower:UpdateTooltip()
-			Called when the widget is hovered. Used to populate the tooltip.
+			Called when mouse is over the widget. Used to populate its tooltip.
```
p3lim: the

elements/alternativepower.lua
```diff
@@ -1,18 +1,17 @@
...
-Handles visibility and updating of the alternative power bar.
-This bar is used to display encounter/quest related power information, such as the number of hour glass uses left on the
-end boss in End Time.
+Handles updating and visibility of a status bar that displays encounter- or quest-related power information, such as the number of hour glass
+charges during Murozond encounter in End Time.
 
 ## Widget
 
-AlternativePower - A StatusBar to represent alternative power.
+AlternativePower - A StatusBar that is used to represent the unit's alternative power.
```
p3lim: drop "that is"

elements/alternativepower.lua
```diff
@@ -1,18 +1,17 @@
...
 
 ## Notes
 
-`OnEnter` and `OnLeave` handlers to display a tooltip will be set on the widget if it is mouse enabled.
-The default StatusBar texture will be applied if the UI widget doesn't have a status bar texture or color defined.
+If mouse interactivity is enabled for the widget, `OnEnter` and/or `OnLeave` handlers will be set to display a tooltip.
+Default status bar texture will be applied if the widget doesn't have a texture or a color set.
```
Rainrider: __A__ default or __The__ default imho

elements/assistantindicator.lua
```diff
@@ -1,15 +1,15 @@
...
+Toggles visibility of an indicator based on the unit's raid assistant status.
 
 ## Widget
 
 AssistantIndicator - Any UI widget.
 
 ## Notes
 
-The default assistant icon will be applied if the UI widget is a texture and doesn't have a texture or color defined.
+Default assistant icon will be applied if the widget is a Texture and doesn't have a texture or a color set.
```
Rainrider: __The__ default. There is only one default, isn't it?

p3lim:
Some notes:
- Castbar has no overrides
- Range has no ForceUpdate method

We also have to decide on the type notation.  
In some cases the documentation prefixes the description of a parameter with the type, e.g. "a Boolean" or "a String", while in others it doesn't.

We need to decide if we want to prefix the descriptions like this or not, and if we do, what format we want to prefix them with ("a Boolean", "boolean value" etc).

p3lim:
- Any reason we arent using the following constants in blizzard.lua?
```
MAX_ARENA_ENEMIES = 5
MAX_BOSS_FRAMES = 5
MAX_PARTY_MEMBERS = 4
```
- The following line doesn't allow for color overriding per-layout, have to override the global colors to change it:
https://github.com/oUF-wow/oUF/blob/development/elements/classpower.lua#L46
- Totems are not hidden while in a vehicle

ls-:
> We need to decide if we want to prefix the descriptions like this or not, and if we do, what format we want to prefix them with ("a Boolean", "boolean value" etc).

IMHO, if we expect to receive parameters of certain types, we usually call `argcheck` to test them, so I think we should mention expected types in such cases, otherwise it's unnecessary.

Moreover, we may want to add `argcheck`s where they're needed, and ofc remove redundant ones.

ls-:
>Totems are not hidden while in a vehicle.
Totems aren't only totems, various CDs, familiars and other things are counted as totems. For instance, DK's Ebon Gargoyle is shown as a "totem".

So theoretically some vehicle may have "totems" too.

p3lim:
> So theoretically some vehicle may have "totems" too.

Has never existed such a thing afaik.

Rainrider:
What about appending the type at the end of the description? Like: `* name - name of the element (string)`
This way the type won't lead to any change in wording or need special highlighting. Lua type could be written in small (i.e. `boolean`), WoW types in PascalCase (i.e. `StatusBar` or `Texture`) (I'd expect anybody reading the docs can keep them apart though). Specifying the type is nice, so people know what to expect.

p3lim:
@Rainrider Good suggestion, and I agree on everything. Singling out widget types is nice, as it's more relevant than variable types.

ls-:
> Has never existed such a thing afaik.

Blizz constantly add/remove abilities to/from "totems", e.g. Mage's Arcane Familiar was a "totem" at some point, now it's just a buff, DK's secondary minion (archer) used to be a "totem" too, but is tied to main summon now.

IMHO, we should leave them as is.

ls-:
@p3lim checked Blizz totem code, although it's shit, they do [this](https://github.com/Gethe/wow-ui-source/blob/live/FrameXML/TotemFrame.lua#L53-L55). So yeah, even theoretically that case is kinda  impossible, guess we can hide it.

Rainrider:
@ls- the code you linked always shows totem timers for shamans and even calculates their placement based on pet presence for DK, so totems actually don't disappear in vehicles.

p3lim:
@Rainrider But the unitframe they are anchored to does. In oUF's case we just swap units on the unitframes, not swapping the entire frames, thus Totems are still connected to the player frame.

ls-:
@Rainrider that's not what I tried to point out :p If `PetFrame` is shown and player's class isn't one of those 6, they just hide `TotemFrame` and prevent it from updating.

@p3lim Default UI does exactly the same thing, when a player gets inside a vehicle `PlayerFrame` shows unit and player is shown as `PetFrame`.

Issue #346, opened 2017-04-22
[7.2.5] Destruction Soul Shards


Gethe:
Per the patch notes:
> Soul Shards are now divided into 10 Soul Shard Fragments for Destruction Warlocks ...

So Soul Shards for Destro Warlocks will be back to what they had in WoD, and would obviously have implications for the class icons element.

Now, while I would be willing to test this, I do not have a leveled Warlock and the PTR doesn't do premades anymore. That said, I'll try to level up a lock on the PTR as I have time.

p3lim:
Should not need any modifications to the element, it works dynamically based on UnitPowerMax.

Gethe:
Ah ok, thanks. Still though, I'd like to be sure.

p3lim:
https://github.com/oUF-wow/oUF/blob/development/elements/classpower.lua#L83

ls-:
I'm just passing by. Code wise nothing changed on PTR, however, there's only one issue, which is quite significant, default shard bars are `StatusBar`s, whereas ours are `Texture`s.

Gethe:
Yes, but you are not using the third argument for that function. Without that, UnitPower[Max] will only return **full** soul shards, in order to get **partial** soul shards you have to pass in true as the third arg.

ls-:
JIC, @Gethe  is talking about [this](https://github.com/Gethe/wow-ui-source/blob/ptr/FrameXML/ShardBar.lua#L103-L107)

```Lua
function WarlockPowerBar_UnitPower(unit)
	local shardPower = UnitPower(unit, Enum.PowerType.SoulShards, true);
	local shardModifier = UnitPowerDisplayMod(Enum.PowerType.SoulShards);
	return (shardModifier ~= 0) and (shardPower / shardModifier) or 0;
end
```

I explained entire situation in our gitter :p

ls-:
@Gethe, yeah, I tested it too... I'd say there's no need to call `UnitPowerMax` w/ 3rd arg, it's easier to do something like this:
```Lua
local full, partial = math.modf(UnitPower("player", powerType, true) / UnitPowerDisplayMod(powerType))
```
ATM we're discussing if we want to convert all our classpowers to `StatusBar`s, jic, Blizz make more class power work this way in the future, or create a separate soulshards/shards/sharbar element.

p3lim:
We'd support both of course, checking the type of the frame/object.

p3lim:
[...]

PR #347, opened 2017-04-22
classpower: Element revamp.

ls-:
The issue was brought up by @Gethe in https://github.com/oUF-wow/oUF/issues/346.

We ended up overhauling entire element 😀🔫

### DIFF
--- a/elements/classpower.lua
+++ b/elements/classpower.lua
@@ -123,6 +123,7 @@ local function Update(self, event, unit, powerType)
 			if(max < oldMax) then
 				for i = max + 1, oldMax do
 					element[i]:Hide()
+					element[i]:SetValue(0)
 				end
 			end
 
--- a/elements/classpower.lua
+++ b/elements/classpower.lua
@@ -83,8 +83,8 @@ local function UpdateColor(element, powerType)
 end
 
 local function Update(self, event, unit, powerType)
-	if(not (unit == 'player' and powerType == ClassPowerType
-		or unit == 'vehicle' and powerType == 'COMBO_POINTS')) then
+	if(not (self.unit == unit and (unit == 'player' and powerType == ClassPowerType
+		or unit == 'vehicle' and powerType == 'COMBO_POINTS'))) then
 		return
 	end
 
--- a/elements/classpower.lua
+++ b/elements/classpower.lua
@@ -60,16 +60,7 @@ local RequireSpec, RequireSpell
 local function UpdateColor(element)
 	local color = oUF.colors.power[ClassPowerType or 'COMBO_POINTS']
 	for i = 1, #element do
-		local icon = element[i]
-		if(element.isTexture) then
-			if(icon.SetDesaturated) then
-				icon:SetDesaturated(PlayerClass ~= 'PRIEST')
-			end
-
-			icon:SetVertexColor(color[1], color[2], color[3])
-		elseif(element.isStatusBar) then
-			icon:SetStatusBarColor(color[1], color[2], color[3])
-		end
+		element[i]:SetStatusBarColor(color[1], color[2], color[3])
 	end
 end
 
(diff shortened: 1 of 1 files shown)


elements/classpower.lua
```diff
@@ -115,7 +148,7 @@ local function Update(self, event, unit, powerType)
 	* event         - the event that triggered the update
 	--]]
 	if(element.PostUpdate) then
-		return element:PostUpdate(cur, max, oldMax ~= max, powerType, event)
+		return element:PostUpdate(full, partial, max, oldMax ~= max, powerType, event)
```
Rainrider: Not related to the commit itself but `event` should be the first argument for consistency reasons, shouldn't it?

elements/classpower.lua
```diff
@@ -243,20 +276,30 @@ local function Enable(self, unit)
...
+				if(icon:IsObjectType('Texture')) then
+					if(not icon:GetTexture()) then
+						icon:SetTexCoord(0.45703125, 0.60546875, 0.44531250, 0.73437500)
+						icon:SetTexture([[Interface\PlayerFrame\Priest-ShadowUI]])
+					end
+
+					areChildrenTextures = true
+				end
 			end
 		end
```
Rainrider: I don't think we should allow the element to hold a mixture of widget types.

elements/classpower.lua
```diff
@@ -265,6 +308,12 @@ local function Enable(self, unit)
 			(element.UpdateTexture or UpdateTexture) (element)
 		end
 
+		if(areChildrenStatusBars) then
+			element.ClassPowerUpdate = UpdateStatusBar
```
Rainrider: The color of the status bars is not set like it is for textures.

elements/classpower.lua
```diff
@@ -56,6 +62,32 @@ local function UpdateTexture(element)
...
+			if(i <= full) then
+				element[i]:SetValue(1)
+			else
+				element[i]:SetValue(partial)
+			end
+		else
+			element[i]:SetValue(0)
+		end
+	end
+end
```
Rainrider: Why not hide empty bars? We do hide the textures, why handle them differently?

p3lim: Runes.

Rainrider: We do not support textures for runes. The update process in this case is inconsistent.

elements/classpower.lua
ls-: Probably, but it's been this way since forever :D

elements/classpower.lua
ls-: Because it'd look sloppy. I (and to be fair not only I) use "smooth" status bars, I want gain and loss animations/processes to be consistent in both cases.

elements/classpower.lua
ls-: https://github.com/oUF-wow/oUF/issues/346#issuecomment-296351867, yeah :p 

Or do you want to make sure that ALL elements in the table are of the same type? I really hope that no addon dev is retarded enough to through few widget types in the mix.

elements/classpower.lua
Rainrider: What if the number of destro soul shards change through a talent or something?
If I don't want to show empty textures, i don't have to do anything layout wise. If I don't want to show empty bars, I'll have to hide them in PostUpdate. If I want to show empty class resources I can use the same background for both bars and textures. So why should one element behave like if it was two elements?

Rainrider: And I'm not sure I understand where the problem is when using smooth bars. Show first, then apply animations.

ls-: Explained everything in gitter w/ gifs and so on.

elements/classpower.lua
```diff
@@ -243,26 +252,27 @@ local function Enable(self, unit)
...
+				isStatusBar = true
 			end
 		end
 
-		if(isChildrenTextures) then
-			--[[ Override: ClassPower:UpdateTexture()
-			Used to completely override the internal function for updating the power icon textures.
+		if(isStatusBar) then
+			--[[ Override: ClassPower:UpdateColor()
+			Used to completely override the internal function for updating the power icon colors.
```
p3lim: Change "power icon colors" to "bars' colors"

elements/classpower.lua
```diff
@@ -243,26 +252,27 @@ local function Enable(self, unit)
 		element.ClassPowerEnable = ClassPowerEnable
 		element.ClassPowerDisable = ClassPowerDisable
 
-		local isChildrenTextures
+		local isStatusBar
 		for i = 1, #element do
 			local icon = element[i]
```
p3lim: Change this variable (and all the following) to `bar`? I personally like `pill`, but w/e.

PR #348, opened 2017-04-24
Power: Check if element is a StatusBar before setting a texture

Gethe:
So my oUF layout in RealUI uses a non-standard status bar. With the current stable version I am able to use the built in update function to without issue, but when testing on the newer dev version ([patch-classpowers](oUF-wow/oUF#347)), I get this error:
``` lua
115x ...terface\nibRealUI\Libs\oUF\elements\power.lua:213: attempt to call method 'SetStatusBarTexture' (a nil value)
...terface\nibRealUI\Libs\oUF\elements\power.lua:213: in function <...terface\nibRealUI\Libs\oUF\elements\power.lua:125>
(tail call): ?
nibRealUI\Libs\oUF\ouf.lua:200: in function <nibRealUI\Libs\oUF\ouf.lua:189>
(tail call): ?
```
The power element already checks if it's actually a status bar in the `Enable` function, so it seemed inconsistent to **not** also check for that in the `Update` function before trying to apply texture settings to it.

With the power element specifically, I think it's fair to assume that it would have :SetValue, and :SetMinMaxValues, but in order to provide flexibility to the layout I think assuming :SetStatusBarTexture and other texture based methods is a step too far.

One of the greatest features of oUF, to me, is it's resilience when given a widget type it does not expect. I would like to see that carried forward in newer versions.

### DIFF
--- a/elements/power.lua
+++ b/elements/power.lua
@@ -122,6 +122,46 @@ local function getDisplayPower(unit)
 	end
 end
 
+local function UpdateColor(element, r, g, b, t)
+	local atlas = element.atlas or (t and t.atlas)
+	if(element.useAtlas and atlas and displayType ~= ALTERNATE_POWER_INDEX) then
+		element:SetStatusBarAtlas(atlas)
+		element:SetStatusBarColor(1, 1, 1)
+
+		if(element.colorTapping or element.colorDisconnected) then
+			t = disconnected and self.colors.disconnected or self.colors.tapped
+			element:GetStatusBarTexture():SetDesaturated(disconnected or tapped)
+		end
+
+		if(t and (r or g or b)) then
+			r, g, b = t[1], t[2], t[3]
+		end
+	else
+		element:SetStatusBarTexture(element.texture)
+
+		if(r or g or b) then
+			element:SetStatusBarColor(r, g, b)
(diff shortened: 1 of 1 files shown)


p3lim:
To be fair, the element has a hard dependency on status bars, it even says so in the top of the file, and the only reason those checks are in the Enable method is so layouts can use the element with an Override when using non-standard widget types.

This use-case is extremely specific, and I'd suggest either overwrite the update function or implement dummy functions for your pseudo widget.

p3lim:
What we are considering, and this has come up before, is splitting out the color parts of Health and Power to their own functions.

That way the color function can be overwritten (in your case to nop), or be called from an overwritten Update method.

Gethe:
OK, that should be fine. Thanks for your consideration.

Gethe:
After a bit of consideration myself, I updated the PR with a possible implementation of that idea. The main goal with this was to maintain the use of the built-in color options, then passing the determined color to `UpdateColor` for they layout to do with what they want.

Rainrider:
We are discussing this internally for the health element as well. My personal opinion is that the whole color related code should be moved away from `Update` to its own function (not only the part depending on the widget type).

PR #349, opened 2017-04-29
castbar: Add support for reverse fill statusbars

Gethe:
`statusbar:SetReverseFill(true)` will make the bar fill from right to left, instead of the default left to right. This commit will adjust `SafeZone` to be on the appropriate side if this attribute is set.


p3lim:
Please rebase this to the development branch and I'll merge it.

PR #355, opened 2017-05-03
range: minor cleanup

Rainrider:
- removed unused ForceUpdate (unneeded because of how the element is updated)
- no need to check for `.Range` in OnUpdate - already done in Enable
- ~~show OnRangeFrame only once~~ rebased
- use `element.insideAlpha` instead of magic 1 in Disable


p3lim:
> show OnRangeFrame only once

It's hidden in Disable.

PR #359, opened 2017-05-10
stagger: Move colour update to its own function

ls-:


elements/stagger.lua
```diff
@@ -94,19 +74,46 @@ local function Update(self, event, unit)
...
+	end
+
+	local cur = UnitStagger('player')
+	local max = UnitHealthMax('player')
+	local perc = cur / max
+
+	element:SetMinMaxValues(0, max)
+	element:SetValue(cur)
+
+	--[[ Override: Stagger:UpdateColor(unit, perc)
```
Rainrider: unit is unused

elements/stagger.lua
```diff
@@ -94,19 +74,46 @@ local function Update(self, event, unit)
...
+	Used to completely override the internal function for updating the widgets' colors.
```
Rainrider: `widget's` instead of `widgets'` - we only have one

elements/stagger.lua
```diff
@@ -94,19 +74,46 @@ local function Update(self, event, unit)
...
+
+	* self - the Stagger element
+	* perc - the percentage of staggered damage (number)
```
Rainrider: Maybe add that it is relative to the player's maximum health.

Rainrider: And maybe add value range

ls-: Why though? It can't be relative to anything else, and it's a percentage, it's 0-100, always...

elements/stagger.lua
ls-: yeah, forgot to remove it  :grin::gun:

elements/stagger.lua
ls-: Copypaste typo o_O
Copied it from health element, it says `widgets'` there...

elements/stagger.lua
Rainrider: It can be relative to what matters to whoever overrides UpdateColor. Also not all ppl think of percentages as values between 0-1.

PR #363, opened 2017-05-16
classpower: Only show active bars

p3lim:
Currently, all bars are shown regardless of current power, and instead fills when the power is available.
This means with e.g. 0 combo points it shows 5 empty bars.

With this patch it shows only those who're (partially) filled.

### DIFF
--- a/elements/classpower.lua
+++ b/elements/classpower.lua
@@ -113,10 +113,17 @@ local function Update(self, event, unit, powerType)
 			mod = UnitPowerDisplayMod(ClassPowerID)
 		end
 
+		local numActive = math.ceil(mod == 0 and 0 or cur / mod)
 		for i = 1, max do
-			element[i]:Show()
-			-- mod should never be 0, but according to Blizz code it can actually happen
-			element[i]:SetValue(mod == 0 and 0 or (cur / mod - i + 1))
+			local bar = element[i]
+			if(i > numActive) then
+				bar:Hide()
+				bar:SetValue(0)
+			else
+				bar:Show()
+				-- mod should never be 0, but according to Blizz code it can actually happen
+				bar:SetValue(mod == 0 and 0 or (cur / mod - i + 1))
+			end
 		end
 
 		oldMax = element.__max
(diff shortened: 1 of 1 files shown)


elements/classpower.lua
```diff
@@ -113,10 +113,17 @@ local function Update(self, event, unit, powerType)
 			mod = UnitPowerDisplayMod(ClassPowerID)
 		end
 
+		local numActive = math.ceil(mod == 0 and 0 or cur / mod)
 		for i = 1, max do
-			element[i]:Show()
-			-- mod should never be 0, but according to Blizz code it can actually happen
-			element[i]:SetValue(mod == 0 and 0 or (cur / mod - i + 1))
+			local bar = element[i]
```
ls-: This one isn't needed, even our own documentation says:
>Define local variables for table references when referring more than twice

ls-:
And it should be more like "classpower: show only somewhat filled bars", anything within [1; max] range is active, but not all active bars are full.

elements/classpower.lua
```diff
@@ -113,10 +113,17 @@ local function Update(self, event, unit, powerType)
 			mod = UnitPowerDisplayMod(ClassPowerID)
 		end
 
+		local numActive = math.ceil(mod == 0 and 0 or cur / mod)
```
Rainrider: You could spare the call to math.ceil by `local numActive = (mod == 0 and 0 or cur / mod) + 0.9`

p3lim:
@ls- When you have 3 out of 5 combo points, then 3 bars are considered "active".
I regret moving this element over to statusbars...

elements/classpower.lua
p3lim: We could, although it's less future proof, what if they introduce a power with a greater accuracy, let's say 100, we'd have to keep adding fractions to that value.

p3lim: Nvm, I just saw the performance difference, math.ceil is 7-10 times slower than that addition!

ls-:
@p3lim I have a bit different opinion on this topic, but whatever :p

PR #386, opened 2017-05-31
Feature: Arena Preperation (take two!)

### LINKED ISSUE #177: Arena preparation frames

p3lim:
This PR supersedes #275 and resolves #177.

By taking over the `arena#` frames' state driver until the unitIDs become available, this PR achieves the following:
- Displays the arena frames with bogus health/power information during arena preperation
    - Has callbacks/overrides and respects `.Override`
- Hides all other (relevant) elements
- Adds Class color support in `[raidcolor]` tag
- Adds new tag `[arenaspec]` which also work on the normal `arena#` frames
- Added `ForceUpdateTags` method (basically exposing an existing function)
- Added `frame:IsEnabled()` shorthand for UnitWatchRegistered

That should cover most of the available features with the arena prep system, and it requires no additional setup for the layout (plug-n-play).

One thing to note, the event `ARENA_PREP_OPPONENT_SPECIALIZATIONS` doesn't seem to fire until a teammate is in the arena with you. I've notified Blizzard about this, as the issue was introduced with 7.0.

Please test!

### DIFF
--- a/ouf.lua
+++ b/ouf.lua
@@ -161,6 +161,14 @@ for k, v in next, {
 		return active and active[name]
 	end,
 
+	--[[ frame:IsEnabled()
+
+	Used to check if the given frame is enabled or not. This is a reference to `UnitWatchRegistered`.
+
+	* self - unit frame
+	--]]
+	IsEnabled = UnitWatchRegistered,
+
 	--[[ frame:Enable(asState)
 	Used to toggle the visibility of a unit frame based on the existence of its unit. This is a reference to
 	`RegisterUnitWatch`.
--- a/ouf.lua
+++ b/ouf.lua
@@ -290,6 +290,13 @@ local function initObject(unit, style, styleFunc, header, ...)
 			else
 				oUF:HandleUnit(object)
 			end
+
+			-- Arena preparation
+			if(unit and unit:match('arena%d?$')) then
+				object:RegisterEvent('ARENA_PREP_OPPONENT_SPECIALIZATIONS', Private.UpdateArenaPreperation, true)
+				-- the event handler only fires for visible frames, so we have to hook it
+				object:HookScript('OnEvent', Private.UpdateArenaPreperation)
+			end
 		else
 			-- Used to update frames when they change position in a group.
 			object:RegisterEvent('GROUP_ROSTER_UPDATE', object.UpdateAllElements)
--- a/arena.lua
+++ b/arena.lua
@@ -2,10 +2,58 @@ local parent, ns = ...
 local oUF = ns.oUF
 local Private = oUF.Private
 
+local function updateElement(self, event, element, specID)
+	local el = self[element]
+	if(el and self:IsElementEnabled(element)) then
+		el:SetMinMaxValues(0, 1)
+		el:SetValue(1)
+
+		local r, g, b, t, _
+		if(el.colorPower and element == 'Power') then
+			-- FIXME: no idea if we can get power type here without the unit
+		elseif(el.colorClass) then
+			local _, _, _, _, _, _, class = GetSpecializationInfoByID(specID)
+			t = self.colors.class[class]
+		elseif(el.colorReaction) then
+			t = self.colors.reaction[2]
+		elseif(el.colorSmooth) then
+			_, _, _, _, _, _, r, g, b = unpack(el.smoothGradient or self.colors.smooth)
+		elseif(el.colorHealth and element == 'Health') then
+			t = self.colors.health
+		end
+
+		if(t) then
+			r, g, b = t[1], t[2], t[3]
+		end
+
+		if(r or g or b) then
+			el:SetStatusBarColor(r, g, b)
+
+			local bg = el.bg
+			if(bg) then
+				local mu = bg.multiplier or 1
+				bg:SetVertexColor(r * mu, g * mu, b * mu)
+			end
+		end
+	end
+end
+
 function Private.UpdateArenaPreperation(self, event)
 	if(event == 'ARENA_OPPONENT_UPDATE' and not self:IsEnabled()) then
 		self:Enable()
+		self:UpdateAllElements('ArenaPreparation')
 		self:UnregisterEvent(event, Private.UpdateArenaPreperation)
+
+		-- show elements that don't handle their own visibility
+		if(self:IsElementEnabled('Auras')) then
+			if(self.Auras) then self.Auras:Show() end
+			if(self.Buffs) then self.Buffs:Show() end
+			if(self.Debuffs) then self.Debuffs:Show() end
+		end
+
+		if(self.Portrait and self:IsElementEnabled('Portrait')) then
+			self.Portrait:Show()
+		end
 	elseif(event == 'PLAYER_ENTERING_WORLD' and not UnitExists(self.unit)) then
 		Private.UpdateArenaPreperation(self, 'ARENA_PREP_OPPONENT_SPECIALIZATIONS')
 	elseif(event == 'ARENA_PREP_OPPONENT_SPECIALIZATIONS') then
@@ -16,7 +64,23 @@ function Private.UpdateArenaPreperation(self, event)
 				self:RegisterEvent('ARENA_OPPONENT_UPDATE', Private.UpdateArenaPreperation)
 			end
 
+			-- pseudo-element updates
+			updateElement(self, event, 'Health', specID)
+			updateElement(self, event, 'Power', specID)
+
+			-- hide all other (relevant) elements
+			if(self.Auras) then self.Auras:Hide() end
+			if(self.Buffs) then self.Buffs:Hide() end
+			if(self.Debuffs) then self.Debuffs:Hide() end
+			if(self.Castbar) then self.Castbar:Hide() end
+			if(self.CombatIndicator) then self.CombatIndicator:Hide() end
+			if(self.GroupRoleIndicator) then self.GroupRoleIndicator:Hide() end
+			if(self.Portrait) then self.Portrait:Hide() end
+			if(self.PvPIndicator) then self.PvPIndicator:Hide() end
+			if(self.RaidTargetIndicator) then self.RaidTargetIndicator:Hide() end
+
 			self:Show()
+			self:ForceUpdateTags()
 		end
 	end
 end
--- a/arena.lua
+++ b/arena.lua
@@ -6,6 +6,23 @@ local function updateElement(self, event, element, specID)
 	local el = self[element]
 	if(el and self:IsElementEnabled(element)) then
 		if(el.Override) then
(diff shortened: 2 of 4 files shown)


ouf.lua
```diff
@@ -161,6 +161,14 @@ for k, v in next, {
 		return active and active[name]
 	end,
 
+	--[[ frame:IsEnabled()
+
+	Used to check if the given frame is enabled or not. This is a reference to `UnitWatchRegistered`.
```
Rainrider: No empty line before the description

Gethe:
So i tested this out and while my arena frames are getting shown during prep, the PostUpdate isn't getting called until the arena actually starts (with event "ArenaPreparation" which seems counter intuitive to me since we are no longer in prep).

This is less than helpful since I am then unable to properly set up the spec icon for my frames during prep.

p3lim:
I'll add some callbacks, thanks for the feedback.

sBaildon:
I had to update the expected return values from GetSpecializationInfoByID(specID) from 7 to 6.

https://github.com/sBaildon/oUF/commit/8172d2548b27798588540bf21cc8b0d25d6ee720

Wutname1:
Is this still planned?

p3lim:
Hopefully, yes, but I don't do arenas.

sBaildon:
I've ran some skirmishes to test this work.

After accounting for [this ](https://github.com/sBaildon/oUF/commit/8172d2548b27798588540bf21cc8b0d25d6ee720) fix, everything works as expected. 

I feel like a callback when the `ARENA_OPPONENT_UPDATE` event finishes firing could be pretty useful

p3lim:
Alright, thanks for testing.
We'll do some refactoring and get this on its way.

p3lim:
Refactoring complete with the following changes:

- Fixed issue with `GetSpecializationInfoByID`
- Added support for UpdateColor on Health and Power elements
- Moved all the arena-specific code to the file we have in space for unit-specific code (derp)

Just need to add callbacks as suggested and it should be good to go.

zorker:
Cannot wait to finally implement arena frames. When will it arrive?!

p3lim:
When it's fully tested.

zorker:
Will add it and do some skirmishes. Need the tokens for recipes right now anyway.

p3lim:
Please test the two tags as well.

ls-:
Yeah, when I had arena frames, it's very similar to boss frames.

zorker:
Ok added them. Will test later. https://github.com/zorker/rothui/commit/59034447b8aa09980c8239d5a4ccf7017d0d3b9f
I will probably add my own tag though. The specName+arenaEnemyName will probably take to much space with the spec included. Displaying the icon via texturestring might be what I will be using.

sBaildon:
> @Gethe & @sBaildon, every frame has a PreUpdate and PostUpdate fired by UpdateAllElements which is also triggered by ARENA_OPPONENT_UPDATE, is that callback not sufficient? The only other things happening after that PostUpdate callback is the showing of Auras/Buffs/Debuffs and Portrait.

I wasn't aware of a global `PostUpdate`, although that makes total sense. My use case is that I want to untag, and then retag some font strings when the real arena frames are spawned—`PostUpdate` should be able to get the job done.

p3lim:
`PostUpdate` should indeed work for that scenario, but make sure to test it, this is the time to get your requests heard.

Gethe:
The global `PreUpdate` and `PostUpdate` callbacks do not work for my use case. As I mentioned in my earlier comment, I'm updating an icon during prep to show the spec. The earliest that these callback are fired is at the start of the match, which is too late to be useful for me.

I currently use the `health:PostUpdateArenaPreparation` callback and this works for my purposes.

I also tested the 'arenaspec' tag and this does work during prep.

zorker:
@Gethe Couldn't you hook the frame OnShow script? In prep mode arena frames will be shown.

Finally home. Will post my results in a bit.

zorker:
Here is my test result: https://imgur.com/a/K8QfjBj

Looks nearly fine. Only one thing is really odd. In prep mode my statusbar only has a slight tint of the class color. Now that I think of it...that totally might be my absorb bar. It is fine as soon as the fight starts.

I will try to change my absorb bar to initialize with a value of 0 and report back.

Gethe:
@zorker That would probably work, but I'd rather not make that hook if there is a sufficient callback to use instead.

zorker:
I tested the OnShow hook. It is working as expected and are available when the prep frames show up.
(That is my OnShow2 print)

While testing I think I found another bug. After winning the arena we started another skirmish without leaving the arena. 

When entering the new arena no prep frames appeared. The arena frames showed asap as the doors opened. The enemy team was monk, paladin, warrior.

https://imgur.com/a/R7MtMom

The updates on the frame once the door opens are fine.

zorker:
Another bug. When leaving the instance normally and joining from the world I get prep frames. But the prep frames do not reset.

In the following instance I had two enemies. It still showed the colors from the enemies from the match before.

https://imgur.com/a/bROixck

Frames where fine once the door was open.

zorker:
I will post all the events I can find here:
https://www.townlong-yak.com/framexml/live/Blizzard_ArenaUI/Blizzard_ArenaUI.xml
https://www.townlong-yak.com/framexml/live/Blizzard_ArenaUI/Blizzard_ArenaUI.lua

The prep stuff is here:
https://www.townlong-yak.com/framexml/live/Blizzard_ArenaUI/Blizzard_ArenaUI.lua#413

Here are the events that I could find:
```
ARENA_PREP_OPPONENT_SPECIALIZATIONS

ARENA_OPPONENT_UPDATE

ARENA_COOLDOWNS_UPDATE
ARENA_CROWD_CONTROL_SPELL_UPDATE

CVAR_UPDATE
VARIABLES_LOADED
PLAYER_ENTERING_WORLD
UNIT_PET
UNIT_NAME_UPDATE
UNIT_HEALTH
UNIT_CLASSIFICATION_CHANGED
```

zorker:
I had some time to think about it. The Blizzard prep frames are unitless frames shown via ARENA_PREP_OPPONENT_SPECIALIZATIONS event that will be displayed via GetArenaOpponentSpec loop.

When ARENA_OPPONENT_UPDATE fires there are the following arguments:
self, event, unit, unitEvent
When unitEvent == "cleared" they hide all the arena prep frames while checking for instanceType arena.
https://www.townlong-yak.com/framexml/live/Blizzard_ArenaUI/Blizzard_ArenaUI.lua#244
Additionally there is a second hide for specific prep frames if a unit GUID was found for the matching id.
https://www.townlong-yak.com/framexml/live/Blizzard_ArenaUI/Blizzard_ArenaUI.lua#148

Since the prep frames are unitless the corresponding self.unit must be set to nil. It might reference to a unit seen before (like in my case where the prep frames showed an old unit).

zorker:
Forget all the above. All the values where fine but still my color did not want get triggered. Turned out it is
https://github.com/oUF-wow/oUF/pull/386/files#diff-e44cc6afad311f577a02c7a4f26a89c6R34

UpdateColor is defined on any Health and Color element. Thus the color code never gets triggered.

zorker:
To be honest the whole UpdateColor part is obsolete. Sadly I could not even work with the Override for the arena prep because that is requiring and Override for the element aswell.

I removed the UpdateColor part alltogether and it is working now. But I had to do some adjustments to my layout to make it look ok.

Here is my arena spawn. I had to hook into preUpdate because I need to reset my absorb bar. I had to hook into PostUpdateArenaPreparation for the power bar because I want it empty and in a bearable color.

```lua
--spawn arena frames
if L.F.CreateArenaStyle then
  oUF:SetActiveStyle(A.."Arena")
  local arena = {}
  local function PostUpdateArenaPreparation(self,event,specID)
    self:SetMinMaxValues(0,1)
    self:SetValue(0)
    local r,g,b = 0.3,0.3,0.3
    self:SetStatusBarColor(r,g,b)
    if self.bg then
      local m = self.bg.multiplier or 0.3
      self.bg:SetVertexColor(r*m,g*m,b*m)
    end
  end
  local function PostUpdate(self,event)
    if event ~= "ARENA_PREP_OPPONENT_SPECIALIZATIONS" then return end
    self.rAbsorbBar:SetMinMaxValues(0,1)
    self.rAbsorbBar:SetValue(0)
  end
  --constant MAX_ARENA_ENEMIES is part of the blizzard arena ui addon which is not loaded on init
  for i = 1, 5 do
    arena[i] = oUF:Spawn("arena"..i, A.."Arena"..i)
    if arena[i].Power then
      arena[i].Power.PostUpdateArenaPreparation = PostUpdateArenaPreparation
    end
    arena[i].PostUpdate = PostUpdate
    if (i == 1) then
      arena[i]:SetPoint(unpack(L.C.arena.point))
    else
      arena[i]:SetPoint(L.C.arena.setup.point, arena[i-1], L.C.arena.setup.relativePoint, L.C.arena.setup.xOffset, L.C.arena.setup.yOffset)
    end
  end
end
```

Suggestion from myself. Remove the UpdateColor part in the unit.lua altogether. On top...why does OverrideArenaPreparation require Override?!

zorker:
Final analysis:

Tag system is working as intended. No issues.

UpdateColor inside updateArenaPreparationElements is bugged. It is defined for Health and Power elements making the actual coloring code inside updateArenaPreparationElements never trigger.
ttps://github.com/oUF-wow/oUF/pull/386/files#diff-e44cc6afad311f577a02c7a4f26a89c6R34

OverrideArenaPreparation  inside updateArenaPreparationElements  requires element.Override and I cannot explain why. One may not want to override the element Update function at all, just the update for the arena prep. They should not rely on each other.
https://github.com/oUF-wow/oUF/pull/386/files#diff-e44cc6afad311f577a02c7a4f26a89c6R10

Power bar coloring does not work. Imo it should to be disabled like Buffs, Auras and Debuffs. It has no use in the prep phase and should be reenabled like Auras when the fight begins. 
https://github.com/oUF-wow/oUF/pull/386/files#diff-e44cc6afad311f577a02c7a4f26a89c6R56
https://github.com/oUF-wow/oUF/pull/386/files#diff-e44cc6afad311f577a02c7a4f26a89c6R111

Here is my curent arena spawn. I had to hook into preUpdate because I need to reset my absorb bar and the debuff highlighting. I would like to get rid of the power bar code, but that requires the power bar to be gone for the prep phase.
https://github.com/zorker/rothui/blob/master/wow8.0/oUF_Simple/core/spawn.lua#L162

Screenshots: https://imgur.com/a/DUblVEj

zorker:
Fair points. 
Tested the commits. Everything is fine now. A go from me.

Gethe:
Yup, from me as well 👍

PR #392, opened 2017-06-17
stagger: add a nil check for UnitStagger

Rainrider:


p3lim:
Please add a comment next to it stating why you need this fallback (BUG or w/e)

Rainrider:
Should I add [this](https://github.com/tomrus88/BlizzardInterfaceCode/blob/master/Interface/AddOns/Blizzard_NamePlates/Blizzard_ClassNameplateBar_Monk.lua#L63-L66) as a comment?

PR #394, opened 2017-06-30
castbar: fix SafeZone width calculation

belzaru17:
safeZonePercent is already multiplied by width below in SetWidth.
self.max is in seconds, and ms is in milliseconds - hence 1e5 is too big.

(safeZonePercent is actually a ratio, but that's not renamed here)

### DIFF
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -91,7 +91,7 @@ local function updateSafeZone(self)
 	-- Guard against GetNetStats returning latencies of 0.
 	if(ms ~= 0) then
 		-- MADNESS!
-		local safeZonePercent = (width / self.max) * (ms / 1e5)
+		local safeZonePercent = (ms / 1e3) / self.max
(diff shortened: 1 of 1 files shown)


belzaru17:
Sure, done. Thanks!

elements/castbar.lua
```diff
@@ -88,19 +88,13 @@ local function updateSafeZone(self)
...
-		safeZone:Show()
-	else
-		safeZone:Hide()
+	local safeZoneRatio = (ms / 1e3) / self.max
+	if(safeZoneRatio > 1) then
+		safeZoneRatio = 1
 	end
+
+	safeZone:SetWidth(width * safeZoneRatio)
+	safeZone:Show()
```
Rainrider: The Show() call should be redundant here. The SafeZone widget should be parented to the Castbar, its visibility is inherited.

belzaru17: Thanks; done.

PR #395, opened 2017-08-20
[WIP] core: Fix nameplate flickering

ls-:
It's a bug in canvas renderer, so I decided to report it instead of creating workarounds.

For info on the bug, see [this thread on WoWI](http://www.wowinterface.com/forums/showthread.php?p=324831).

### DIFF
--- a/ouf.lua
+++ b/ouf.lua
@@ -708,9 +708,15 @@ function oUF:SpawnNamePlates(namePrefix, nameplateCallback, nameplateCVars)
 		eventHandler:RegisterEvent('PLAYER_LOGIN')
 	end
 
-	eventHandler:SetScript("OnUpdate", function()
-		for blizz, ouf in next, nameplates do
-			ouf:SetAlpha(blizz:GetAlpha())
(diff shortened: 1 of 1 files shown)


ouf.lua
```diff
@@ -707,6 +708,18 @@ function oUF:SpawnNamePlates(namePrefix, nameplateCallback, nameplateCVars)
 		eventHandler:RegisterEvent('PLAYER_LOGIN')
 	end
 
+	eventHandler:SetScript('OnUpdate', function(handler, elapsed)
+		handler.elapsed = (handler.elapsed or 0) + elapsed
+
+		if handler.elapsed >= 0.05 then
```
p3lim: Missing parentheses, maybe good to add a note here that it matches 60 fps?

p3lim:
I understand the need for it, although I don't particularly like the solution, even if it's the only one that works.
We'll hopefully get to review this later if some fixes are applied on Blizzard's end.

ls-:
Decided to report it instead of creating workarounds.

PR #399, opened 2017-10-14
Fix portrait updates

ls-:
WIP.


Rainrider:
Reported on [wowi](http://www.wowinterface.com/forums/showthread.php?p=325461) (for the record)

Issue #402, opened 2017-11-28
Possible issue on new boss Antoran High Command (vehicle override)


Blazeflack:
Hi.

We have received a report from a user of ElvUI that people who enter the vehicles (Legion Command Pod) during the new Antoran High Command raid boss, have their raid frame turned into a vehicle, which he is unable to target in the usual left-click way. He has to use a /target command instead. I'm not raiding or even actively playing WoW myself, so I am unable to confirm or debug this properly.

While I understand that there are times where the layout is at fault, especially with the kind of modifications ElvUI makes, I thought I would put out a feeler here just in case. There have been some changes in oUF recently regarding vehicles, so it's not too far off to think that this could be an unintentional side-effect.

If any of you guys happen to notice anything, or if you can confirm that your own layout works just fine on that boss, then please let me know here.

Thank you.

p3lim:
Are you using oUF 7.x in ElvUI now?

Blazeflack:
Hi p3lim.

Yes we are up-to-date with latest changes from the master branch here.

p3lim:
Alright, I don't raid myself so I can't really test it, but it seems unintentional indeed.

Blazeflack:
I appreciate your input, thank you. The raid just opened this week so hopefully other people will be able to confirm or deny it soon-ish.

Grimsbain:
I raid using a normal oUF raid layout and it is indeed messing up on the pods. When the person gets in their heal stops updating.

Blazeflack:
Thanks for the confirmation @Grimsbain

Grimsbain:
I don't know enough about the backend side of oUF to suggest what code needs to be fixed but the [changes to CompactUnitFrame_UpdateInVehicle](https://www.townlong-yak.com/framexml/24920/CompactUnitFrame.lua/diff) in 7.3 may be a clue to what is needed.

p3lim:
Leaving the original ticket from ElvUI here too.
https://git.tukui.org/elvui/elvui/issues/220

Blizzard seems to have flip-flopped the vehicle logic, we'll investigate further and get a fix out soon™.

p3lim:
Can any of you confirm that players in pods are targetable with the default raid frames?

By the looks of the default implementation, Blizzard seems to have disabled all vehicle logic completely.

ls-:
[...]

PR #403, opened 2017-11-30
core: set the `toggleForVehicle` attribute to false

Rainrider:
Blizzard has changed the way vehicles work for Antoran High Command (see #402)
This way vehicles are still targetable through the corresponding pet buttons (see http://wowwiki.wikia.com/wiki/SecureActionButtonTemplate#Other_attributes)

### DIFF
--- a/ouf.lua
+++ b/ouf.lua
@@ -274,7 +274,8 @@ local function initObject(unit, style, styleFunc, header, ...)
 
 			-- No need to enable this for *target frames.
 			if(not (unit:match('target') or suffix == 'target')) then
-				object:SetAttribute('toggleForVehicle', true)
+				-- BUG: Blizzard has changed the way vehicles work for Antoran High Command
+				object:SetAttribute('toggleForVehicle', false)
 			end
 
 			-- Other boss and target units are handled by :HandleUnit().
@@ -550,7 +551,8 @@ do
 
 				frame:SetAttribute('*type1', 'target')
 				frame:SetAttribute('*type2', 'togglemenu')
-				frame:SetAttribute('toggleForVehicle', true)
+				-- BUG: Blizzard has changed the way vehicles work for Antoran High Command
+				frame:SetAttribute('toggleForVehicle', false)
 				frame:SetAttribute('oUF-guessUnit', unit)
 			end
 


pquerner:
Shouldnt you check if you are in the Antorus raid for this ? Or does it apply for other vehicle fights aswell. (Malygos, Ulduar some bosses, ....)

Rainrider:
The change means that the pet button (unit frame) will represent the vehicle and the player button will represent the player. Same goes for party and raid meaning that as long as the layout supports party/raid pets the user has the ability to see and target the vehicles.

Malygos and Ulduar are solo content, so you don't even need to change anything as long as your layout supports the player pet frame. You can even use `/tar pet` to target your vehicle. So in this regard it is more of a cosmetic change than anything else.

Apart from that we are limited in when we can change secure attributes. I'm not sure that your proposal is feasible because of that. However Ulduar is about to get an update for timewalking, so I suppose Blizzard will look into this until then.

Rainrider:
Btw one side effect of this is that combo points for the 3rd phase of the Malygos encounter or the Aces High daily quest are not being shown anymore. This could be worked around and the stock ui fails there too.

p3lim:
> Shouldnt you check if you are in the Antorus raid for this ? Or does it apply for other vehicle fights aswell. (Malygos, Ulduar some bosses, ....)
This sadly breaks all targetting of vehicles in raid frames, but oUF is intended to implement everything as close to stock as possible, and we'll revert this when Blizzard does.

p3lim:
Please test this in Antorus, I don't raid so I can't do it.

ouf.lua
```diff
@@ -274,7 +274,8 @@ local function initObject(unit, style, styleFunc, header, ...)
 
 			-- No need to enable this for *target frames.
 			if(not (unit:match('target') or suffix == 'target')) then
-				object:SetAttribute('toggleForVehicle', true)
+				-- BUG: Blizzard has changed the way vehicles work for Antoran High Command
+				object:SetAttribute('toggleForVehicle', false)
```
p3lim: Let the non-header units have vehicle support, this issue only concerns vehicles.

ls-:
Honestly, I'm not sure about not swapping units on player and playerpet frames.

I understand that it's done to fix the issue for people who're manually creating 40 raid frames instead of using headers, but it in some cases messes up layouts, e.g., it tries to show combo points for classes/specs that don't have any, thus texturing and stuff may be off.

People who aren't using headers are minority, people who have player and playerpet frames are majority, making life easier for people who purposefully decided to make it as hard as possible (people w/ non-header raid frames) at the expense of majority of userbase is not cool.

p3lim:
> Honestly, I'm not sure about not swapping units on player and playerpet frames.

That is not the intention, probably just a bit overzealous with the editing.

pquerner:
> Malygos and Ulduar are solo content

There is achievements to do (atleast kill one Boss) Ulduar with Level 80 and iLVL 370 (or something). This change would break for those players.

In the Malygos fight, you actually heal the dragon, not the player. This change would break for players who do the raid in the original style (level 85?) aswell.

The list goes on, ... 

So this "temp" fix (if it really is ever fixed by Blizz) should be locked down better tbh. (Like checking for Level 110, if on Raidboss in Antorus, ...)

p3lim:
@pquerner The stock UI behaves the same.

> So this "temp" fix (if it really is ever fixed by Blizz) should be locked down better tbh. (Like checking for Level 110, if on Raidboss in Antorus, ...)

Some of this is not possible due to secure environment restrictions.

ls-:
I'm testing another solution now... >_>

ouf.lua
Rainrider: This does not remove vehicle support, it changes the way it works.

Resike:
Don't get me wrong but if the healer has their _raid header_ and _raidpet header_ shown too, then when someone enters a vehicle that player should be targetable in the _raidpet header_ as a player, and in the _raid header_ as a vehcile?

p3lim:
@Resike That was the behaviour before 7.3.2, yes.

pquerner:
As I installed this patch I started lagging at the Portal Boss in Antorus. I dont know if that fix could cause this. (Lagging as in every other second my fps goes down to 1fps, then up again after ~3sec)

p3lim:
See #404

PR #404, opened 2017-12-01
core: Disable raid frames vehicle handling in Antorus raid

ls-:
It's WIP fix for #402, unlike #403 it doesn't break raid vehicle frames in other raids like Ulduar or EoE.

Logic is quite straightforward. When you zone into Antorus, it'll hack raid frames, but when you zone out it'll unhack them. If you're in combat, this un/hackery will be delayed :grin::gun:

@pquerner, @Blazeflack, @Grimsbain test it, pls 😭 

Suggestions are also welcome...

### DIFF
--- a/ouf.lua
+++ b/ouf.lua
@@ -508,7 +508,7 @@ do
 	end
 
 	-- There has to be an easier way to do this.
-	local initialConfigFunction = [[
+	local initialConfigFunctionTemp = [[
 		local header = self:GetParent()
 		local frames = table.new()
 		table.insert(frames, self)
@@ -550,7 +550,7 @@ do
 
 				frame:SetAttribute('*type1', 'target')
 				frame:SetAttribute('*type2', 'togglemenu')
-				frame:SetAttribute('toggleForVehicle', true)
+				frame:SetAttribute('toggleForVehicle', %d == 1)
 				frame:SetAttribute('oUF-guessUnit', unit)
 			end
 
@@ -568,6 +568,9 @@ do
 			clique:RunAttribute('clickcast_register')
 		end
 	]]
+
+	-- it's needed for vehicle hack
+	local initialConfigFunction = initialConfigFunctionTemp:format(1)
 
 	--[[ oUF:SpawnHeader(overrideName, template, visibility, ...)
 	Used to create a group header and apply the currently active style to it.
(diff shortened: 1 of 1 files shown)


Merathilis:
I wonder if the map id is correct. If i run: `/run print(GetMapNameByID(1712))` it returns nothing, so i assume the correct map id is 1188

Blazeflack:
I'm going to release a version of ElvUI with the fix from #403 just so raiders have something that is working with current content at least. I can't personally test the method from this pull request, but if it proves successful then I will add this to ElvUI instead of the other workaround.

Thanks.

ls-:
@Merathilis, I was about to tag you :D

1712 is not a map ID, it's an instance map ID T_T

Merathilis:
@ls- Oh, i assume i need moar coffee :>

Grimsbain:
I tested it using the Eye of Eternity raid (switched the instance id) and it seems to work fine.

ls-:
@Grimsbain, it's fine, no need to hurry :p

ls-:
Fixed loading logic, now it should work w/ layouts that use oUF's factory method.

@Rainrider, ty for testing <3

Rainrider:
It's fugly but it works. It leaves singly spawned group frames out though, so layouts that don't use headers will have to either adapt or take #403, which in turn has issues with old content if the layout does not spawn group pets, leading to ppl being unable to target group members' vehicles through oUF.

pquerner:
I am having trouble installing this over ElvUIs version. Cannot test it this ID as the boss is killed. :/

ls-:
From my testing in ToC (id: 650) it works fine :grin::gun:

I used ElvUI and @Rainrider's layout for testing.

Merathilis:
I have pull your recent changes inside of ElvUI: https://git.tukui.org/elvui/elvui/tree/oUF-Vehicle_fix

But i will be able to test this probably next week in my main raid.

p3lim:
@Merathilis If ElvUI is not using headers you need to fetch again.
Specifically for https://github.com/oUF-wow/oUF/pull/404/commits/ae5569b70f7b8957e2b655ab7382112166e397ab

Merathilis:
Sry... have missed that commit, thank you very much @p3lim

p3lim:
Keep in mind that we strongly advice against using single units for groups, and they don't have the full benefit from this PR, they won't get toggled on/off based on location, like headers do.

Merathilis:
A litte feedback for your fix, i was able to test it on normal (with ElvUI and your latest changes from the "hackery branch".

So far, it was working fine for me.:

unbenannt

Also the target frame seems to be fine tho: 
unbenannt1

p3lim:
Thanks for testing @Merathilis.

Blazeflack:
Thanks @p3lim but ElvUI uses headers for group frames :)

p3lim:
Issued a warning on WoWInterface:
http://www.wowinterface.com/forums/showthread.php?t=55893

pquerner:
@Merathilis What happend to the branch?

Issue #410, opened 2017-12-29
Possible issue with unit handling on arena frames in PvP Brawls


Blazeflack:
Someone on our team noticed a possible issue with unit handling on arena frames in PvP Brawls. The arena frames are used to show flag carriers, one from each faction. He noticed that clicking an arena frame didn't always target the unit that was being displayed on the frame.

Someone else was able to confirm that it seems to work correctly with the default Blizzard frames. It only shows 2 flag carriers (even though there can be 6 in the brawl). It shows whoever picked up a flag first, and then switches to someone else if/when the first flag carrier drops the flag.

The usual disclaimer applies here, as I can't guarantee that the issue is with oUF and not just ElvUI due to us not using the :Spawn functionality in oUF.

If any of the people who watches this project does PvP on occasion then I would appreciate your insights.

Thank you.

Cutepally:
hi, im simpy. sorry this post is just a copy/paste of conversation but it may hold some useful (idk) information.
for reference: https://git.tukui.org/elvui/elvui/commit/1a33d01c this is a temp fix for ElvUI to just disable the frames inside of this brawl, that i did last night. not directly related to any oUF code.

```
Resike - Today at 5:55 AM
i've run a brawl yesterday and the default blizz frames seemed to work for me properly
there is 2 fc frames, whoever picks the flag up first, then jumps to the second fc when someone dies
```
```
Simpy - Today at 9:35 AM
o hmm
@Resike on blizz ui (0 addons) i had no arena frames for flag carrier
so wut
ill check out that copychat thing tho blaze
Resike - Today at 9:42 AM
yeah i had that just now too
but i had no issues with targeting
Simpy - Today at 9:42 AM
mine would target but it would have the wrong color/class/name
target/health matched
but the info on it was wrong for me
Resike - Today at 9:42 AM
thats an usual arena bug
Simpy - Today at 9:43 AM
sigh
i never had problems with these in reg bgs but this one is goofed for me >x>
Resike - Today at 9:44 AM
it has something to do with if your party is not full then the unit id's are fucked up
```
```
Simpy - Today at 9:47 AM
hmmm
tbh im not sure
all i know is
i was chain running that brawl
every game they were broken
and i got tired of seeing my name and the health go down lol thinking im dying or something when i was fine XD
Resike - Today at 9:51 AM
isnt ouf unregisters them?
Simpy - Today at 9:52 AM
thats handled by us and what i "fixed"
we enable the frames always
if we have them on
[...]

PR #411, opened 2017-12-31
core: update arena frames on ARENA_OPPONENT_UPDATE

Rainrider:
Fixes https://github.com/oUF-wow/oUF/issues/410


p3lim:
Happy New Year's! 🎆🎉

PR #414, opened 2018-01-03
changelog: Add links to issues mentioned in commit messages

p3lim:


haste:
This should just match for `#\d+` and use that for GitHub issue linking. Seems odd to only allow `\(#\d+\)$`.

Rainrider:
@haste There might be other numbers in the commit message. Or do you mean something else?

haste:
Other numbers in the commit message in the format of `#<digits>`?

Rainrider:
Ah, yes, the `#`

Issue #409, opened 2017-12-25
strange implementations in health prediction


selfaddicted:
https://github.com/oUF-wow/oUF/blob/5a6cf9701b6c9368f0b669d44a3d7aae2b8555ee/elements/healthprediction.lua#L108-L112

Replacing the value of healAbsorb with that of Health at the line would bias the calculations after the line. healAbsorb Bar's max value should not be over health bar's length but shouldn't it be done later?

Rainrider:
Flying back back later today, will take at look later.

Rainrider:
https://github.com/tomrus88/BlizzardInterfaceCode/blob/37e971677afa43e2d2a6cd39dff2e34aeaedf180/Interface/FrameXML/UnitFrame.lua#L273-L279 reference to stock ui code

The healAbsorb is placed above the the health bar because healing someone with a healabsorb would not change their actual health value but deplete the healabsorb first. This is the default ui behavior and it seems logical to me.

Yet, I'm a bit out of context right now, so feel free to object.

selfaddicted:
first of all, I'm not English speaker, so please consider it :)

Let's say an example, a unit whose current health is 1000 and got a debuff of heal absorb 2000, and  there is incoming heal as 2400 which can erase the debuff.

so the logic should be like this; the incoming heal covers the heal absorb and the remained 400 incoming heal should be shown. BUT, the code overrides the heal absorb before considering incoming heal, finally, the incoming heal would be 2400 - 1000 = 1400, which isn't right.

Am I right?

Rainrider:
If you want to test your assumptions, run the following code on https://www.lua.org/cgi-bin/demo and adjust the values accordingly.
This is basically the body of the update function of the healthprediction element. `overflow` is a substitution for `element.maxOverflow`

```lua
local overflow = 1
local myIncomingHeal = 2400
local allIncomingHeal = 2400
local absorb = 0
local healAbsorb = 2000
local health, maxHealth = 1000, 5000

local hasOverHealAbsorb = false
if(health < healAbsorb) then
	hasOverHealAbsorb = true
	healAbsorb = health
end

if(health - healAbsorb + allIncomingHeal > maxHealth * overflow) then
	allIncomingHeal = maxHealth * overflow - health + healAbsorb
end

local otherIncomingHeal = 0
if(allIncomingHeal < myIncomingHeal) then
	myIncomingHeal = allIncomingHeal
else
[...]

PR #415, opened 2018-01-06
[healthprediction] Update maths [WIP]

ls-:
Fixes maths related to heal absorb and incoming heals values. Previously we're discarding the excess of heal absorbs, while it should've been used to adjust inc heals values.

~I'll also update element's docs tomorrow, an example of `PostUpdate` is required to show people how to properly imitate Blizz prediction bars' behaviour. In our case `absorbBar` needs to be dynamically appended to either health bar or `otherBar`, depending on which one is the rightmost.~

The main difference is that inc heals and heal absorbs are mutually exclusive now. If the amount of heal absorb is larger than the amount of inc heals, its value will be adjusted and inc heals will be discarded. and vice versa.

This issue was highlighted in #409.

### DIFF
--- a/elements/healthprediction.lua
+++ b/elements/healthprediction.lua
@@ -137,7 +137,7 @@ local function Update(self, event, unit)
 			hasOverAbsorb = true
 		end
 
-		absorb = math.max(0, maxHealth - (health + allIncomingHeal))
+		absorb = math.max(0, maxHealth - health - allIncomingHeal)
 	end
 
 	if(element.myBar) then
--- a/elements/healthprediction.lua
+++ b/elements/healthprediction.lua
@@ -132,7 +132,7 @@ local function Update(self, event, unit)
 	end
 
 	local hasOverAbsorb = false
-	if(health + allIncomingHeal + absorb >= maxHealth or health + absorb >= maxHealth) then
+	if(health + allIncomingHeal + absorb >= maxHealth) then
 		if(absorb > 0) then
 			hasOverAbsorb = true
 		end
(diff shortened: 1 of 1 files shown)


ls-:
Also, I may add one more arg for `PostUpdate` to show that inc heals are there, but they get absorbed, so devs could do something w/ it.

ls-:
Meh, turns out we don't want to have mandatory `PostUpdate`s in oUF, so I'll rewrite it tomorrow.

ls-:
The main difference is that inc heals and heal absorbs are mutually exclusive now. If the amount of heal absorb is larger than the amount of inc heals, its value will be adjusted and inc heals will be discarded. and vice versa.

This way PostUpdate-voodoo isn't needed.

elements/healthprediction.lua
```diff
@@ -104,41 +104,40 @@ local function Update(self, event, unit)
...
+			otherIncomingHeal = allIncomingHeal - myIncomingHeal
 		end
 	end
 
-	if(healAbsorb > allIncomingHeal) then
-		healAbsorb = healAbsorb - allIncomingHeal
-	else
-		healAbsorb = 0
+	local hasOverAbsorb = false
+	if(health + allIncomingHeal + absorb >= maxHealth or health + absorb >= maxHealth) then
```
Rainrider: The `health + absorb >= maxHealth` condition is not needed. `allIncomingHeal` is always non-negative.

ls-:
> Maybe add a note that incoming heals adjustments for heal absorb are at the cost of others' heals first and player's heals after that.

IMHO, not needed, it's kinda obvious from the code.

PR #418, opened 2018-01-15
Auras: PostUpdateIcon update

Rainrider:
This adds duration, expiration, dispelType and isStealable to the list of PostUpdateIcon arguments.

`duration` and `expiration` are needed by layouts for custom aura timers. This would spare them the need to call UnitAura again or use CustomFilter to access those values.

`dispelType` and `isStealable` are interesting because the auras element hides the aura border (.overlay) when the .showType/.showBuffType/.showDebuffType are not found, which is a strange behavior in the first place. So this is a way to allow layouts to react to it without breaking backwards compatibility.

### DIFF
--- a/elements/auras.lua
+++ b/elements/auras.lua
@@ -198,7 +198,7 @@ local function updateIcon(element, unit, index, offset, filter, isDebuff, visibl
 
 			if(button.overlay) then
 				if((isDebuff and element.showDebuffType) or (not isDebuff and element.showBuffType) or element.showType) then
-					local color = DebuffTypeColor[dispelType] or DebuffTypeColor.none
+					local color = DebuffTypeColor[dispelType or 'none']
 
 					button.overlay:SetVertexColor(color.r, color.g, color.b)
 					button.overlay:Show()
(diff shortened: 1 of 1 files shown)


elements/auras.lua
```diff
@@ -228,14 +228,18 @@ local function updateIcon(element, unit, index, offset, filter, isDebuff, visibl
...
-			* index    - the index of the aura (number)
-			* position - the actual position of the aura button (number)
+			* self        - the widget holding the aura buttons
+			* unit        - the unit on which the aura is cast (string)
+			* button      - the updated aura button (Button)
+			* index       - the index of the aura (number)
+			* position    - the actual position of the aura button (number)
+			* duration    - the aura duration in seconds (number?)
+			* expiration  - the point in time when the aura will expire. Comparable to GetTime() (number)
+			* dispelType  - the dispel type of the aura (string?)['Curse', 'Disease', 'Magic', 'Poison']
```
p3lim: Blizzard refers to this as "debuffType".

Rainrider: `dispelType` is used in the code, so I just kept it that way. Do you want me to change just the docs or also rename the variable?

PR #422, opened 2018-01-26
Header names

Rainrider:
Makes generated names for headers a bit closer to their intended purpose. There are still cases where layouts are better off providing global names themselves.

### DIFF
--- a/ouf.lua
+++ b/ouf.lua
@@ -454,9 +454,9 @@ local function generateName(unit, ...)
 	for i = 1, select('#', ...), 2 do
 		local att, val = select(i, ...)
 		if(att == 'showRaid') then
-			raid = true
+			raid = val ~= false and val ~= nil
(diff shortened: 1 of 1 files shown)


p3lim:
Could you do a TL;DR before and after examples of the effect of this PR?

Rainrider:
Commenting in the commit order:
1. `showRaid, false` is a valid attribute - value pair for a party header. The old behavior would name it raid in that case.
2. both TANK and MAINTANK are valid header filters with different meanings, the original code was written before class roles were added
3. fixes wrong names for i.e. maintank/mainassist targets headers
4. Not sure why I did this right now, probably found the numbers misleading and thought being more verbose was a good idea
5. consistency change (see the target unit suffix)

p3lim:
Messy commit messages but otherwise LGTM.

PR #424, opened 2018-01-28
core: Update raid vehicle handling hack

### LINKED ISSUE #423: Intermittent Issue on Antoran High Command
We are seeing a few reports that some people experience the issue with the vehicle pods on the Antoran High Command once more. The weird part is that the ElvUI team is unable to reproduce the issue.

One user provided info that disabling all addons except ElvUI worked for him, but after he re-enabled all his addons again it was still working correctly. To me this sounds like that it may not be an addon issue, but rather something that might get fixed with a /reload or a relog.

Do you guys have any thoughts about whether the currently implemented workaround in oUF might fail to register correctly upon entering the raid instance? That's the only thing I can think of that might be the cause, and would also explain why a /reload would fix it.

Edit: The ElvUI ticket for reference: https://git.tukui.org/elvui/elvui/issues/279

ls-:
Supposedly fixes #423 :grin::gun:

@Blazeflack, @Merathilis, hopefully, this one will work better :D

### DIFF
--- a/ouf.lua
+++ b/ouf.lua
@@ -648,6 +648,9 @@ do
 
 	-- The remainder of this scope is a temporary fix for issue #404,
 	-- regarding vehicle support on headers for the Antorus raid instance.
+	-- Track changes to SecureButton_GetModifiedUnit, this hack should be
+	-- removed when UnitTargetsVehicleInRaidUI is added to it. Supposedly,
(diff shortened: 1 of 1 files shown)


ls-:
Technically, `ZONE_CHANGED_NEW_AREA` block is no longer needed, but I left it there just in case :p

PR #430, opened 2018-04-15
Update wowprogramming.com links in documentation

jukx:
The (very helpful!) links to wowprogramming.com in the inline documentation are currently throwing 404s. This can lead to quite some confusion (and did for me...). The fix was to just append ".html".


p3lim:
Seems like an error with the webserver on their end, I'll try and get in contact with them.

haste:
I've poked jnwhiteh about this on IRC.

jukx:
The *.html links are used on [the site itself](http://wowprogramming.com/docs/secure_template.html) by the way. So it might be an intended change. Still a bit weird.

PR #443, opened 2018-07-24
Add the tag [runes] to display the amount of runes (player only)

Industrial:


Rainrider:
You should add the event at which the tag is supposed to update (RUNE_POWER_UPDATE).
Please keep the coding style consistent:
- braces around if conditions (no space between _if_ and the opening brace)
- spaces around binary operators (= in this case)

No need to explicitly check `if ready == true`, `if(ready)` is enough in this case.

p3lim:
@Industrial still interested on getting this through?

ls-:
Welp, I've just merged it into another branch to do the cleanup there, I wasn't sure if you reply here or not :D

PR #451, opened 2018-08-27
auras: prevent cooldown spiral from darkening count

sBaildon:
This PR turns this:
countframe2

Into this:
countframe

### DIFF
--- a/elements/auras.lua
+++ b/elements/auras.lua
@@ -96,8 +96,10 @@ local function createAuraIcon(element, index)
 	local icon = button:CreateTexture(nil, 'BORDER')
 	icon:SetAllPoints()
 
-	local count = button:CreateFontString(nil, 'OVERLAY', 'NumberFontNormal')
-	count:SetPoint('BOTTOMRIGHT', button, 'BOTTOMRIGHT', -1, 0)
+	local countFrame = CreateFrame('Frame', '$parentCountFrame', button)
+	countFrame:SetAllPoints(button)
+	local count = countFrame:CreateFontString(nil, 'OVERLAY', 'NumberFontNormal')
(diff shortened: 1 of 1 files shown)


ls-:
It's unnecessary and should be handled by the layout.

p3lim:
I wouldn't call it unnecessary, although I don't particularily like the implmentation ([I do the same myself](https://github.com/p3lim-wow/oUF_P3lim/blob/master/oUF_P3lim.lua#L201-L210)).

Have you tried playing around with layers/strata? I haven't dealt with this particular issue in a long time, things might have changed since.

ls-:
I do the same thing as well, but I still think it should be handled by the layout. I won't object if you decide to add though.

Also `frame:SetFrameLevel(button.cd:GetFrameLevel() + 1)` is better than some hardcoded frame lvl.

ls-:
K, you convinced me :D

Then you need to change few things:
- Add an empty line between `countFrame` and `count` blocks;
- Add `countFrame:SetFrameLevel(cd:GetFrameLevel() + 1)` to make sure that proper frame levels are assigned, sometimes these things are a bit wonky, so better safe than sorry;
- Remove `countFrame` name, imho, it's unnecessary, so `nil` it.

@p3lim any suggestions?

sBaildon:
New commit for your requests!

Rainrider:
Is the creation of an extra frame to hold the count font string necessary? Can't we use `cd:CreateFonrString(...)`?

p3lim:
@Rainrider No, because of two reasons:

1. It makes the count dependant on cooldowns, some people outright disables them
2. Cooldowns are hidden/shown automatically; https://github.com/oUF-wow/oUF/blob/94c18c96403a2dd1251211d01bf575848e3b3679/elements/auras.lua#L194-L201

PR #463, opened 2018-09-25
core: Remove remaining old vehicle hack code

ls-:


Rainrider:
Left-over from #404 after ebed1e3fd48fd65d2073534daf3e35db60d74273 (just for reference)

PR #467, opened 2018-10-22
Add oUF:UnitSelectionColor(unit) method and .colorSelection option

ls-:
Blizz UI no longer uses `FACTION_BAR_COLORS` for unit frames' or nameplates' name colouring, instead, it uses `UnitSelectionColor` method. It's been this way since Legion.

`oUF:UnitSelectionColor(unit)` is a wrapper for `UnitSelectionColor` that allows us to use custom colours from the `oUF.colors.reacation` table.

~Technically, we could replace `.colorReputation` w/ this, but more options = more better, ayy\~ :grin::gun:~

No docs yet.

-- update 1

~K, since other colours have been discovered, reusing `oUF.colors.reaction` is no longer reasonable, so I'll create a new table for it.~

~List of colours I haven't added to the wrapper function yet:~
~- `255, 255, 139` - the colour for your own char while in combat;~
~- `128, 128, 128` - the colour for dead units.~

-- update 2

Rewrote the whole thing, still no docs, trying to catch all the colours.

"Selection" is a character's outline and a circle underneath, this method returns its colour.

### DIFF
--- a/colors.lua
+++ b/colors.lua
@@ -2,6 +2,8 @@ local parent, ns = ...
 local oUF = ns.oUF
 local Private = oUF.Private
 
+local print = Private.print
+
 local frame_metatable = Private.frame_metatable
 
 local colors = {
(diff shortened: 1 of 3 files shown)


ls-:
Not going to merge it in its current state, it's too hacky for oUF standards.

Let's hope that Blizz devs agree to improve selection API in the future.

Issue #426, opened 2018-02-12
ClassPower backgrounds aren't shown


kavasto:
The statusbars are hidden if the power point is inactive, so the bg is never visible.

Rainrider:
Don't parent your backgrounds to the bars then?

kavasto:
I'm already working around this with PostUpdate. Just reporting here because it seems odd to use statusbars and then hide them instead of just relying on the value. Maybe it does make sense and I'm just missing something, though.

zorker:
Afaik the classbar is only hidden when the current power value is 0. Which is especially useful for rogues. You only see the combo point class bar if you have any. I find it convenient that the classbar hides itself if you have no points in it.

I use my own classbar module in oUF_Simple because I use a different element approach but I do it the same way. When the current power value is 0 the bar will hide.

Rainrider:
A persistent background is something we consider a decoration, which means it is up to the layout. The whole element historically used textures, but because of desto locks, we switched to statusbars for simplicity. We kept the behavior like it was with textures, where you had to layer and show your backgrounds a decorations yourself. The .bg option came to it because it is something common and oUF provides it for statusbars in other elements across the board.

PR #468, opened 2018-11-26
Always show combo point backgrounds

ajprax:
Hiding classpower bars when inactive overrides the desired behavior of showing the "empty" bar background meaning that the empty bars only show up on login/reload. After the first time Update is called, they're always either active or entirely invisible.

### DIFF
--- a/elements/classpower.lua
+++ b/elements/classpower.lua
@@ -133,11 +133,13 @@ local function Update(self, event, unit, powerType)
 
 		oldMax = element.__max
 		if(max ~= oldMax) then
-			for i = max + 1, oldMax do
-				if(max < oldMax) then
+			if(max < oldMax) then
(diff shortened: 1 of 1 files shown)


ajprax:
Tested when gaining and losing combo points (entering and leaving cat form) and when changing max combo points (specing into and out of Deeper Stratagem).

Rainrider:
The reason for the current design was https://github.com/oUF-wow/oUF/issues/426#issuecomment-365189171. When we did the switch from textures to statusbars, it meant layouts had to only update the widget type to accommodate for it (depending on their implementation ofc), so basically your PR is changing the elements behavior for current layouts.

Since we are about to have a new major release soon (tm), we might reconsider it though.

Rainrider:
Sorry, we decided against the PR because of the aforementioned reasons.

PR #470, opened 2018-12-09
Add missing tag events

Rainrider:
Sorted the tagEvents table alphabetically, which kinda obfuscates the first commit, sorry for that


elements/tags.lua
```diff
@@ -265,17 +265,12 @@ local tagStrings = {
...
-			cp = GetComboPoints('vehicle', 'target')
-		else
-			cp = GetComboPoints('player', 'target')
-		end
+		local cp = UnitPower(u, Enum.PowerType.ComboPoints)
 
 		if(cp > 0) then
 			return cp
 		end
-	end]],
```
p3lim: You removed too much.

Rainrider: thanks mate

PR #471, opened 2018-12-12
New SummonIndicator element

ls-:
It's added with 8.1, the name is self-explanatory. I'll add docs a bit later, a bit busy w/ work IRL >_>

I also sorted elements in oUF.xml alphabetically :grin::gun:


elements/summonindicator.lua
```diff
@@ -0,0 +1,92 @@
...
+local SUMMON_STATUS_ACCEPTED = Enum.SummonStatus.Accepted or 2
+local SUMMON_STATUS_DECLINED = Enum.SummonStatus.Declined or 3
+
+local function Update(self)
+	local element = self.SummonIndicator
+
+	--[[ Callback: SummonIndicator:PreUpdate()
+	Called before the element has been updated.
+
+	* self - the Health element
```
Rainrider: copy/pasta? :)

elements/summonindicator.lua
```diff
@@ -0,0 +1,92 @@
...
+
+	* self  - the parent object
+	* event - the event triggering the update (string)
+	* ...   - the arguments accompanying the event
+	--]]
+	return (self.SummonIndicator.Override or Update) (self, ...)
+end
+
+local function ForceUpdate(element)
+	if(not element.__owner.unit) then return end
```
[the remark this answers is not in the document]
ls-: It's copypasta from raidtargetindicator.lua :V

elements/summonindicator.lua
```diff
@@ -0,0 +1,92 @@
...
+	return Path(element.__owner, 'ForceUpdate')
+end
+
+local function Enable(self)
+	local element = self.SummonIndicator
+	if(element) then
+		element.__owner = self
+		element.ForceUpdate = ForceUpdate
+
+		self:RegisterEvent('INCOMING_SUMMON_CHANGED', Path, true)
```
Rainrider: ISC is not a unitless event according to the docs: https://github.com/tomrus88/BlizzardInterfaceCode/blob/master/Interface/AddOns/Blizzard_APIDocumentation/UnitDocumentation.lua#L304

However it won't work, because our event code is smarter than us: https://github.com/oUF-wow/oUF/blob/master/events.lua#L80-L83 (or add the event to https://github.com/oUF-wow/oUF/blob/master/events.lua#L15 but I think this is easy to get out of sync). Btw, the last time I checked, you could register unitless events with units just fine (frame:RegisterUnitEvent('SPELLS_CHANGED', 'player') worked), so maybe tell our event code to not take decisions on the user's account.

PR #472, opened 2018-12-18
Patch unit events

Rainrider:
Basically, deciding silently for the layout which event is unitless is bad.
Events like PARTY_MEMBER_ENABLE, INCOMING_RESURRECT_CHANGED, PLAYER_SPECIALIZATION_CHANGED and others are all unit events but were always wrongly considered unitless.

#### Side notes:
The following code from a layout:
```lua
self:RegisterEvent('PLAYER_TOTEM_UPDATE', Update)
```
will produce the following error:
```
3x oUF\events.lua:98: Attempt to register unknown event "PLAYER_TOTEM_UPDATE"
[C]: ?
oUF\events.lua:98: in function `RegisterEvent'
oUF_Layout\elements\custom\totems.lua:75: in function `enable'
oUF\ouf-@project-version@.lua:113: in function `EnableElement'
oUF\ouf-@project-version@.lua:329: in function <oUF\ouf.lua:251>
(tail call): ?
oUF\ouf-@project-version@.lua:685: in function `Spawn'
oUF_Layout\layout.lua:7: in function `func'
oUF\factory.lua:20: in function <oUF\factory.lua:16>
(tail call): ?
...
```

This is because PLAYER_TOTEM_UPDATE is unitless and cannot be registered as a unit event. Blame Blizz about the wording of the error message.

This is a **breaking change**. Layouts will have to mark unitless events by using the 4th param of RegisterEvent.

The possibility of mixed registration may be a thing. All event handlers should check if they got the expected unit.

### DIFF
--- a/private.lua
+++ b/private.lua
@@ -24,3 +24,17 @@ end
 function Private.UnitExists(unit)
 	return unit and (UnitExists(unit) or ShowBossFrameWhenUninteractable(unit))
 end
+
+local validator = CreateFrame('Frame')
+
+function Private.validateUnit(unit)
+	local isOK, _ = pcall(validator.RegisterUnitEvent, validator, 'UNIT_HEALTH', unit)
+	if isOK then
+		_, unit = validator:IsEventRegistered('UNIT_HEALTH')
+		if unit then
+			validator:UnregisterEvent('UNIT_HEALTH')
+
+			return true
+		end
+	end
+end
--- a/elements/assistantindicator.lua
+++ b/elements/assistantindicator.lua
@@ -27,6 +27,7 @@ local oUF = ns.oUF
 
 local function Update(self, event)
 	local element = self.AssistantIndicator
+	local unit = self.unit
 
 	--[[ Callback: AssistantIndicator:PreUpdate()
 	Called before the element has been updated.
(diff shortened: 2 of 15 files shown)


Rainrider:
I think this is working as intended now. Please test and comment.

To sum it up:
- Unitless events have to be explicitly marked as such when calling oUF's `RegisterEvent`. This is not backwards-compatible and thus a **breaking change**.
- Blizzard's `RegisterUnitEvent` behaves strangely and falls back silently to the behavior of `RegisterEvent` when passing explicit `nil` for either of the unit arguments. This is also true if only invalid units were passed to it.
- Unit event registration for header units has to be postponed until the frame units are known.
- In order to support mixed event registration (unit events can be registered either for specific units or for all), handlers are NOT guarantied to be only run for the frame units. Thus, checking the event unit against the frame unit is advised (this is not new however).

private.lua
```diff
@@ -24,3 +24,17 @@ end
...
 	return unit and (UnitExists(unit) or ShowBossFrameWhenUninteractable(unit))
 end
+
+local validator = CreateFrame('Frame')
+
+function Private.validateUnit(unit)
+	local isOK, _ = pcall(validator.RegisterUnitEvent, validator, 'UNIT_HEALTH', unit)
+	if isOK then
+		_, unit = validator:IsEventRegistered('UNIT_HEALTH')
+		if unit then
```
p3lim: Missing paranteses in this and the above conditional.

PR #479, opened 2019-01-19
core: allow elements to be overridden

Azilroka:
This will help with debugging custom elements vs default elements without having to change the name each time.


Rainrider:
What is the issue with naming your element differently? Your change would actually make debugging harder, because a custom element by the same name would silently override the default.

PR #480, opened 2019-01-28
castbar: Element revamp

ls-:
The purpose of this PR is to revamp the castbar element.

It's been largely untouched for almost a decade, there's a number of features missing, and the amount of code duplication is mind-boggling.

The first goal is to get rid of excessive code duplication by merging events' handlers like so:

- `UNIT_SPELLCAST_START` + `UNIT_SPELLCAST_CHANNEL_START` => `CastStart`
- `UNIT_SPELLCAST_DELAYED` + `UNIT_SPELLCAST_CHANNEL_UPDATE` => `CastUpdate`
- `UNIT_SPELLCAST_STOP` + `UNIT_SPELLCAST_CHANNEL_STOP` => `CastStop`
- `UNIT_SPELLCAST_INTERRUPTED` + `UNIT_SPELLCAST_FAILED` => `CastFail`
- `UNIT_SPELLCAST_NOT_INTERRUPTIBLE` + `UNIT_SPELLCAST_INTERRUPTIBLE` => `CastInterruptible`

Changes:

Because a good half of events' handlers was removed, some of callbacks were removed as well, others were renamed and had their args changed, so the list of new callbacks looks like this:

- `element:PostCastStart(unit, name)` + `element:PostChannelStart(unit, name)`  
  => `element:PostCastStart(unit)`

- `element:PostCastDelayed(unit, name)` + `element:PostChannelUpdate(unit, name)`  
  => `element:PostCastUpdate(unit)`

- `element:PostCastStop(unit)` + `element:PostChannelStop(unit)`
  => `element:PostCastStop(unit, spellID)`

- `element:PostCastFailed(unit)` + `element:PostCastInterrupted(unit)`  
  => `element:PostCastFail(unit, spellID)`

- `element:PostCastNotInterruptible(unit)` + `element:PostCastInterruptible(unit)`  
  => `element:PostCastInterruptible(unit)`

Added `spellID` to `PostCastStop` and `PostCastFail` since they may be useful to some, but at that point they're no longer available via `element.spellID`.

From now on positioning of the `.Spark` sub-widget should be handled by the layout, for instance:
```lua
Spark:SetPoint("CENTER", Castbar:GetStatusBarTexture(), "RIGHT", 0, 0)
```

`.SafeZone` now properly supports vertically-oriented cast bars.

It's no longer possible to create castbars for targettarget, focustarget, and other similar units since their respective unit frames aren't driven by events, but rely on the `OnUpdate` script.

Added the `.hideTradeSkills` option, the name is self-explanatory, added it since the same option is present in the default UI.

### DIFF
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -395,7 +395,7 @@ local function Enable(self, unit)
 
 		element:SetScript('OnUpdate', element.OnUpdate or onUpdate)
 
-		if(self.unit == 'player' and not (self.hasChildren or self.isChild)) then
+		if(self.unit == 'player' and not (self.hasChildren or self.isChild or self.isNamePlate)) then
 			CastingBarFrame_SetUnit(CastingBarFrame, nil)
 			CastingBarFrame_SetUnit(PetCastingBarFrame, nil)
 		end
@@ -443,7 +443,7 @@ local function Disable(self)
 
 		element:SetScript('OnUpdate', nil)
 
-		if(self.unit == 'player' and not (self.hasChildren or self.isChild)) then
+		if(self.unit == 'player' and not (self.hasChildren or self.isChild or self.isNamePlate)) then
 			CastingBarFrame_OnLoad(CastingBarFrame, 'player', true, false)
 			PetCastingBarFrame_OnLoad(PetCastingBarFrame)
 		end
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -391,6 +391,8 @@ local function Enable(self, unit)
 		self:RegisterEvent('UNIT_SPELLCAST_INTERRUPTIBLE', CastInterruptible)
 		self:RegisterEvent('UNIT_SPELLCAST_NOT_INTERRUPTIBLE', CastInterruptible)
 
+		element.holdTime = 0
+
 		element:SetScript('OnUpdate', element.OnUpdate or onUpdate)
 
 		if(self.unit == 'player' and not (self.hasChildren or self.isChild)) then
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -22,17 +22,17 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
 
 ## Options
 
-.timeToHold      - indicates for how many seconds the castbar should be visible after a _FAILED or _INTERRUPTED
+.timeToHold      - Indicates for how many seconds the castbar should be visible after a _FAILED or _INTERRUPTED
                    event. Defaults to 0 (number)
-.hideTradeSkills - makes the element ignore casts related to crafting professions (boolean)
+.hideTradeSkills - Makes the element ignore casts related to crafting professions (boolean)
 
 ## Attributes
 
-.castID           - a globally unique identifier of the currently cast spell (string?)
-.casting          - indicates whether the current spell is an ordinary cast (boolean)
-.channeling       - indicates whether the current spell is a channeled cast (boolean)
-.notInterruptible - indicates whether the current spell is interruptible (boolean)
-.spellID          - the spell identifier of the currently cast/channeled spell (number)
+.castID           - A globally unique identifier of the currently cast spell (string?)
+.casting          - Indicates whether the current spell is an ordinary cast (boolean)
+.channeling       - Indicates whether the current spell is a channeled cast (boolean)
+.notInterruptible - Indicates whether the current spell is interruptible (boolean)
+.spellID          - The spell identifier of the currently cast/channeled spell (number)
 
 ## Examples
 
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -393,7 +393,7 @@ local function Enable(self, unit)
 
 		element:SetScript('OnUpdate', element.OnUpdate or onUpdate)
 
-		if(self.unit == 'player') then
+		if(self.unit == 'player' and not (self.hasChildren or self.isChild)) then
 			CastingBarFrame_SetUnit(CastingBarFrame, nil)
 			CastingBarFrame_SetUnit(PetCastingBarFrame, nil)
 		end
@@ -441,7 +441,7 @@ local function Disable(self)
 
 		element:SetScript('OnUpdate', nil)
 
-		if(self.unit == 'player') then
+		if(self.unit == 'player' and not (self.hasChildren or self.isChild)) then
 			CastingBarFrame_OnLoad(CastingBarFrame, 'player', true, false)
 			PetCastingBarFrame_OnLoad(PetCastingBarFrame)
 		end
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -23,7 +23,7 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
 
 .timeToHold      - indicates for how many seconds the castbar should be visible after a _FAILED or _INTERRUPTED
                    event. Defaults to 0 (number)
-.hideTradeSkills - makes the element ignore casts related to crafted professions (boolean)
+.hideTradeSkills - makes the element ignore casts related to crafting professions (boolean)
 
 ## Attributes
 
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -344,15 +344,14 @@ local function onUpdate(self, elapsed)
 		self:SetValue(duration)
 
 		if(self.Spark) then
-			local horiz = self.horizontal
-			local size = self[horiz and 'GetWidth' or 'GetHeight'](self)
-
+			local isHoriz = self.isHorizontal
+			local size = self[isHoriz and 'GetWidth' or 'GetHeight'](self)
 			local offset = (duration / self.max) * size
-			if(self:GetReverseFill()) then
+			if(self.isReversed) then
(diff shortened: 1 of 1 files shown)


elements/castbar.lua
```diff
@@ -83,443 +85,290 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
...
+	else
+		element.duration = endTime - GetTime()
+	end
+
+	element:SetMinMaxValues(0, element.max)
 	element:SetValue(0)
 
-	if(element.Text) then element.Text:SetText(text) end
-	if(element.Icon) then element.Icon:SetTexture(texture) end
+	if(element.Icon) then element.Icon:SetTexture(texture or [[Interface\ICONS\INV_Misc_QuestionMark]]) end
```
p3lim: Use a constant for the fallback texture

elements/castbar.lua
```diff
@@ -83,443 +85,290 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
...
+			safeZone:SetPoint(element:GetReverseFill() and (isHoriz and 'LEFT' or 'BOTTOM') or (isHoriz and 'RIGHT' or 'TOP'))
+		else
+			safeZone:SetPoint(element:GetReverseFill() and (isHoriz and 'RIGHT' or 'TOP') or (isHoriz and 'LEFT' or 'BOTTOM'))
+		end
 
-	local element = self.Castbar
-	if(element.castID ~= castID) then
-		return
-	end
+		local ratio = (select(4, GetNetStats()) / 1000) / element.max
```
p3lim: Avoid using `select`

p3lim:
The only breaking change that I could see is the callback name change.
If you'd add a compatibility layer for the old ones, we don't necessarily have to bump major.

elements/castbar.lua
ls-: IMHO, it should be fine in this case, bit I'll remove it if you insist T_T

elements/castbar.lua
ls-: Thought so, even did so initially, but... >_>

p3lim (dismissed):
LGTM, but as this has no actual benefit other than cleaner code, we'll wait with merging this until the next major bump, rather than dealing with backwards compatibility and deprecation.

elements/castbar.lua
```diff
@@ -83,443 +84,292 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
...
-local function UNIT_SPELLCAST_STOP(self, event, unit, castID)
-	if(self.unit ~= unit and self.realUnit ~= unit) then return end
+	-- Channeled spells for some reason don't have castIDs
+	if(element.castID ~= castID or element.spellID ~= spellID) then return end
 
-	local element = self.Castbar
-	if(element.castID ~= castID) then
-		return
-	end
+	if(element.Spark) then element.Spark:Hide() end
```
Rainrider: This might be needed in `CastFail` because of the `.timeToHold` option, but not here or do I miss something?

elements/castbar.lua
```diff
@@ -83,443 +84,292 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
...
 
-local function UNIT_SPELLCAST_CHANNEL_START(self, event, unit, _, spellID)
-	if(self.unit ~= unit and self.realUnit ~= unit) then return end
+local function CastFail(self, event, unit, castID, spellID)
+	if(self.unit ~= unit) then return end
 
 	local element = self.Castbar
-	local name, _, texture, startTime, endTime, _, notInterruptible = UnitChannelInfo(unit)
-	if(not name) then
+	if(not element:IsShown() or element.castID ~= castID or element.spellID ~= spellID) then
```
Rainrider: Same as above

elements/castbar.lua
```diff
@@ -51,6 +51,7 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
     local Spark = Castbar:CreateTexture(nil, 'OVERLAY')
     Spark:SetSize(20, 20)
     Spark:SetBlendMode('ADD')
+    Spark:SetPoint('CENTER', Castbar:GetStatusBarTexture(), 'RIGHT', 0, 0)
```
Rainrider: This one is not documented as a sub-widget.

elements/castbar.lua
```diff
@@ -83,443 +84,292 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
...
 end
 
-local function UNIT_SPELLCAST_INTERRUPTED(self, event, unit, castID)
-	if(self.unit ~= unit and self.realUnit ~= unit) then return end
+local function CastUpdate(self, event, unit, castID, spellID)
+	if(self.unit ~= unit) then return end
 
 	local element = self.Castbar
-	if(element.castID ~= castID) then
+	if(not element:IsShown() or element.castID ~= castID or element.spellID ~= spellID) then
```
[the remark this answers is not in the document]
Rainrider: Btw is the `element:IsShown()` check really needed?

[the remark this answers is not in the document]
ls-: The `:IsShown()` checks are needed since those events fire for crafting spells as well, in those cases both `castID` and `spellID` will be nil. Looks like I actually missed one in `CastStop`

The absence of `castID`s in channels' events' payloads is the reason why I compare `spellID`s instead, we don't want to accidentally updated a bar for some other channelled ability. It's unlikely to happen, but better safe than sorry.

And Blizz code hasn't been updated after they reworked events' payloads, so it's a bad reference.

This comment applies to all of your comments as well.

elements/castbar.lua
ls-: 👌

elements/castbar.lua
ls-: There's no need to show the spark, when the bar is being set to its max value. It applies to both fails and stops, even the default castbars behave similarly.

elements/castbar.lua
```diff
@@ -83,443 +85,289 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
...
-	local sf = element.SafeZone
-	if(sf) then
-		sf:ClearAllPoints()
-		sf:SetPoint(element:GetReverseFill() and 'RIGHT' or 'LEFT')
-		sf:SetPoint('TOP')
-		sf:SetPoint('BOTTOM')
-		updateSafeZone(element)
-	end
+	resetAttributes(element)
+	element:SetValue(element.max)
```
p3lim: Should be set to 0 if channeling.

elements/castbar.lua
```diff
@@ -83,443 +85,289 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
...
 
-	element:SetMinMaxValues(0, max)
+	if(element.casting) then
+		element.duration = GetTime() - startTime
+	else
+		element.duration = endTime - GetTime()
+	end
+
+	element:SetMinMaxValues(0, element.max)
 	element:SetValue(0)
```
p3lim: Should be set to `element.max` if channeling.

ls-:
Rebased to resolve conflicts w/ the master branch.

ls-:
Rebased to remove conflicts with the shadowlands branch.

elements/castbar.lua
```diff
@@ -83,462 +85,289 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
...
 	end
 
-	element.casting = nil
-	element.notInterruptible = nil
-	element.holdTime = element.timeToHold or 0
-
-	--[[ Callback: Castbar:PostCastFailed(unit)
-	Called after the element has been updated upon a failed spell cast.
+	--[[ Callback: Castbar:PostCastStart(unit)
+	Called after the element has been updated upon a spell cast start.
```
Rainrider, line 172: Called after the element has been updated upon a spell cast**/channel** start.

elements/castbar.lua
```diff
@@ -83,462 +85,289 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
...
 	end
 
-	element.casting = nil
-	element.notInterruptible = nil
+	resetAttributes(element)
 
-	--[[ Callback: Castbar:PostCastStop(unit)
-	Called after the element has been updated when a spell cast has finished.
+	--[[ Callback: Castbar:PostCastStop(unit, spellID)
+	Called after the element has been updated when a spell cast has stopped.
```
Rainrider, line 248: Called after the element has been updated when a spell cast**/channel** has stopped.

elements/castbar.lua
```diff
@@ -83,462 +85,289 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
...
-		positionSafeZone(element, true)
-		updateSafeZone(element)
-	end
+	resetAttributes(element)
+	element:SetValue(element.max)
 
-	--[[ Callback: Castbar:PostChannelStart(unit, name)
-	Called after the element has been updated upon a spell channel start.
+	--[[ Callback: Castbar:PostCastFail(unit, spellID)
+	Called after the element has been updated upon a failed spell cast.
```
Rainrider, line 279: Called after the element has been updated upon a failed**/interrupted** spell cast.

Rainrider (commented):
Only some clarification in the docs. Other than that LGTM

Rainrider (approved):
LGTM

PR #490, opened 2019-03-15
Update the colouring of elements

ls-:
The main goal of this PR is to separate colour updates from value updates where possible. So instead of being bundled up as one big `Update` function that handles both maths/behaviour and colouring, there will be two functions:
- `Update` for maths and behaviour
- `UpdateColor` for colouring

This way users will be able to override these individually. Why would we want that? Because, at least from my experience, layout devs tend to override our `Update` functions to alter or completely disable the colouring.

In addition to that, new colours and colour-related options will be added to oUF and its elements. For instance, threat and alternative power colours weren't present in the `oUF.colors` table.

### DIFF
--- a/elements/alternativepower.lua
+++ b/elements/alternativepower.lua
@@ -13,6 +13,32 @@ AlternativePower - A `StatusBar` used to represent the unit's alternative power.
 If mouse interactivity is enabled for the widget, `OnEnter` and/or `OnLeave` handlers will be set to display a tooltip.
 A default texture will be applied if the widget is a StatusBar and doesn't have a texture set.
 
+## Options
+
+.useAtlas                         - Use this to let the widget use an atlas for its texture if an atlas is present in
+                                    `self.colors.power` for the alternative power type (boolean)
+.smoothGradient                   - 9 color values to be used with the .colorSmooth option (table)
+.considerSelectionInCombatHostile - Indicates whether selection should be considered hostile while the unit is in
+                                    combat with the player (boolean)
+
+The following options are listed by priority. The first check that returns true decides the color of the bar.
+
+.colorThreat       - Use `self.colors.threat[threat]` to color the bar based on the unit's threat status. `threat` is
+                     defined by the first return of [UnitThreatSituation](https://wow.gamepedia.com/API_UnitThreatSituation) (boolean)
+.colorPower        - Use `self.colors.power[token]` to color the bar based on the unit's alternative power type
+                     (boolean)
+.colorClass        - Use `self.colors.class[class]` to color the bar based on unit class. `class` is defined by the
+                     second return of [UnitClass](http://wowprogramming.com/docs/api/UnitClass.html) (boolean)
+.colorClassNPC     - Use `self.colors.class[class]` to color the bar if the unit is a NPC (boolean)
+.colorSelection    - Use `self.colors.selection[selection]` to color the bar based on the unit's selection color.
+                     `selection` is defined by the return value of Private.unitSelectionType, a wrapper function
+                     for [UnitSelectionType](https://wow.gamepedia.com/API_UnitSelectionType) (boolean)
+.colorReaction     - Use `self.colors.reaction[reaction]` to color the bar based on the player's reaction towards the
+                     unit. `reaction` is defined by the return value of
+                     [UnitReaction](http://wowprogramming.com/docs/api/UnitReaction.html) (boolean)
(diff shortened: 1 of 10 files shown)


elements/runes.lua
```diff
@@ -80,27 +80,46 @@ local function descSort(runeAID, runeBID)
...
-local function UpdateColor(element, runeID)
+local function UpdateColor(self, event)
+	local element = self.Runes
+
 	local spec = GetSpecialization() or 0
 
 	local color
-	if(spec > 0 and spec < 4 and element.colorSpec) then
-		color = element.__owner.colors.runes[spec]
+	if(spec  > 0 and spec < 4 and element.colorSpec) then
```
Rainrider, line 89: One extra space here :)
Also, tbh, I don't think we need to keep the `.colorSpec` option as Blizzard's defaults are now coloring by spec and there are no more different rune types for the last two expansions.

Rainrider (changes requested):
Runes colour is not applied on /reload.

elements/alternativepower.lua
```diff
@@ -50,9 +80,60 @@ local function onLeave()
...
+	if(element.colorThreat and not UnitPlayerControlled(unit) and UnitThreatSituation('player', unit)) then
+		t =  self.colors.threat[UnitThreatSituation('player', unit)]
+	elseif(element.colorPower) then
+		t = self.colors.power[ALTERNATE_POWER_INDEX]
+
+		if(element.useAtlas and t and t.atlas) then
+			atlas = t.atlas
+		end
+	elseif(element.colorClass and UnitIsPlayer(unit)) or
+		(element.colorClassNPC and not UnitIsPlayer(unit)) then
```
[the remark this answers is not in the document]
ls-: I copypasted it from `health.lua` without any reformatting, so it's been this way for ages, at least a decade. I'll fix trailing `and`s and `or`s in other elements as well.

elements/runes.lua
ls-: Yeah, I saw it too, just didn't want to have a separate commit for it :D

ls-:
> Runes colour is not applied on /reload.
@Rainrider it happens because `Path` and `ColorPath` are completely independent for runes, there's no overlapping events, so `ColorPath` or `ColorUpdate` are never called from `Path`. And `Path` is used as the main function for forced updates from `UpdateAllElements`. I'm going to pass `ForceUpdate` instead of `Path` to `AddElement`, and, tbhon, I wouldn't mind doing so for all other elements as well, but probably in a separate PR.

Rainrider:
@ls- apparently I forgot to update the deprecated and removed `UnitAlternatePowerInfo`. Could you change `getDisplayPower` in the power element to:
```lua
local function getDisplayPower(unit)
	local barInfo = GetUnitPowerBarInfo(unit)
	if(barInfo and barInfo.showOnRaid) then
		return ALTERNATE_POWER_INDEX, barInfo.minPower
	end
end
```
alternatively we may want
```lua
local function getDisplayPower(unit)
	local barInfo = GetUnitPowerBarInfo(unit)
	if(barInfo and (unit == 'player' or barInfo.showOnRaid)) then
		return ALTERNATE_POWER_INDEX, barInfo.minPower
	end
end
```
so that alt power is always shown for the player with the `.displayAltPower` option.

I could also push this myself but don't know if you have any pending changes on the branch.

Rainrider:
@ls- The classpower element is missing `PostUpdateColor` and `PostUpdateColor` is missing documentation.

ls-:
@Rainrider classpower wasn't missing it locally, I just hadn't pushed it :D And I'm reluctant to write docs before all changes are finalised.

ls-:
@Rainrider as for alt power, that option is pretty much exclusively for raid/party frames, it's not useful for anything else. So I'll leave it as `barInfo and barInfo.showOnRaid`.

Rainrider:
WoWScrnShot_092720_203801
This is Onyxia in  Blackwing Descent. The power element will fail to display the alt power in place of the normal one on the boss frame. I actually use `power.displayAltPower` for boss units only.

ls-:
@Rainrider tbhon, it's a misuse because most bosses with alt power use both power AND alt power.

ls-:
But if you insist, then I'll remove all, but `barInfo` checks.

Rainrider:
All raid members with alt power have power at the same time :) Leave it like that then, I can always override it in my layout.

Rainrider (approved):
LGTM

Rainrider (approved):

PR #494, opened 2019-03-31
runes: Fixed sorting

ls-:
This typo I made ages ago was brought to my attention today :grin::gun:


ls-:
Let's just pretend that the current implementation is how it's meant to be, and just mention it in docs 🤣:gun:

PR #501, opened 2019-05-02
tags: Update the system

ls-:
This PR primarily is targetted at layout devs who have in-game configs. Its goal is to improve the process of creating custom tags via GUIs.

Normally, when you write a new tag in Lua, you can just store additional constants, tables, or helper methods in the higher level of scope, and then directly access those from your tag code. However, you can't really do that when your tag is just a string that's loaded via `loadstring`. Previously, you'd have to either create additional tables or even funcs inside the tag string, and those would be created on every single tag func call, or create global vars, which isn't optimal at all.

So, let's say you want to create a tag named `"tag"` that returns `"yes!"` if your target's class is either mage or shaman, then your tag's code will look like this:

```lua
local canPurge = {
	MAGE = true,
	SHAMAN = true,
}

oUF.Tags.Methods["tag"] = function()
	return canPurge[UnitClass('target')] and "yes!" or "no!"
end
```

However, you can't access the local `canPurge` table if your tag's code is a string:

```lua
oUF.Tags.Methods["tag"] = [[function()
	return canPurge[UnitClass("target")] and "yes!" or "no!"
end]]
```

That's where the new `oUF.Tags.Vars` comes in.

```Lua
oUF.Tags.Vars["tag"] = {
	canPurge = {
		MAGE = true,
		SHAMAN = true,
	},
}
```

Now you can access the `canPurge` table from at `_VARS["tag"].canPurge`:

```lua
oUF.Tags.Methods["tag"] = [[function()
	return _VARS["tag"].canPurge[UnitClass("target")] and "yes!" or "no!"
end]]
```

If you're good at naming vars and/or aren't afraid of name collisions, you can do something like this:

```Lua
oUF.Tags.Vars["canPurge"] = {
	MAGE = true,
	SHAMAN = true,
}

oUF.Tags.Methods["tag"] = [[function()
	return _VARS.canPurge[UnitClass("target")] and "yes!" or "no!"
end]]
```

`_VARS` is shared by all oUF tags, so it might be better to keep certain things in their own tables, but things that are used by multiple tags could be stored in `_VARS` as is.

`oUF.Tags.Vars` accepts all types of vars, however, strings are a special case. oUF will try to parse a string w/ `loadstring`, if possible, execute it, and store its result, but in case of failure, it'll just store that string as is.

-- edit 

Added two new methods to oUF.Tags: `:RefreshMethods(tag)` and `:RefreshEvents(tag)`.

`:RefreshMethods(tag)` regenerates/rebuilds tag functions for all tag strings that use provided tag, allows for tag functions' updates w/o reloading the UI.

`:RefreshEvents(tag)` unregisters and then registers events on font strings that use provided tag, this method will allow layout devs to reregister events after editing event lists w/o using `Untag` + `Tag` methods.

### DIFF
--- a/elements/tags.lua
+++ b/elements/tags.lua
@@ -601,6 +601,7 @@ end
 
 local tagPool = {}
 local funcPool = {}
+local fsPool = {}
 local tmp = {}
 
 local function getTagName(tag)
@@ -827,6 +828,7 @@ local function Tag(self, fs, tagstr, ...)
 		end
 	end
 
+	fsPool[fs] = tagstr
 	self.__tags[fs] = true
 end
 
@@ -850,6 +852,7 @@ local function Untag(self, fs)
 
 	fs.UpdateTag = nil
 
+	fsPool[fs] = nil
 	self.__tags[fs] = nil
 end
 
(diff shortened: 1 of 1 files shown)


ls-:
Added two new methods to oUF.Tags: `:Recache(tag) and `:ReregisterEvents(tag)`. Names are PHs, will come up w/ better ones later, I hope 😭

`:Recache(tag)` regenerates/rebuilds tag functions for all tag strings that use provided tag, allows for tag functions' updates w/o reloading the UI.

`:ReregisterEvents(tag)` unregisters and then registers events on font strings that use provided tag, this method will allow layout devs to reregister events after editing event lists w/o using `Untag` + `Tag` methods.

PR #511, opened 2019-07-09
auras: Add .tooltipAnchor option

ls-:
Due to [8.2 UI changes](https://us.forums.blizzard.com/en/wow/t/ui-changes-in-rise-of-azshara/202487), we should avoid parenting GameTooltip to frames with anchoring restrictions, otherwise, GameTooltip will inherit said restrictions.

This PR addresses #510, but it also adds an option to adjust GameTooltip's anchor point w/o the need to using the `CreateIcon` override.

### DIFF
--- a/elements/auras.lua
+++ b/elements/auras.lua
@@ -28,6 +28,8 @@ At least one of the above widgets must be present for the element to work.
 .initialAnchor      - Anchor point for the icons. Defaults to 'BOTTOMLEFT' (string)
 .filter             - Custom filter list for auras to display. Defaults to 'HELPFUL' for buffs and 'HARMFUL' for
                       debuffs (string)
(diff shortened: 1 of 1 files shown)


p3lim:
The anchoring restriction checks should have comments on them, for later reference.

Wutname1:
@ls-  This fix does not seem to be working for me. It is anchoring to the buff button even when manually specifying cursor.  When i add a print to self:GetParent().__owner.tooltipAnchor it returns nil.

Looks like "self:GetParent().__owner" is the nameplate frame it's self and not the Auras frame.
Calling self:GetParent().__owner.Auras.tooltipAnchor works but obviously not a real fix as it only works in 1/3rd of scenarios.

ls-:
Damn,  yeah, `self:GetParent().__owner.tooltipAnchor` should be `self:GetParent().tooltipAnchor` >_>

I'll PR a fix either later today or tomorrow, I'll add comments @p3lim asked for as well.

PR #512, opened 2019-07-18
Fix .tooltipAnchor option and add info about restrictions

ls-:
Ref: #511


Wutname1:
I added ca23531ed624dc193735c926f2dec698bdbfb4f2 to my code earlier and can confirm it's working as intended now.

ls-:
Yeah, I originally had it that way, but while cleaning things up, I accidentally copied that `__owner` bit from `UpdateTooltip` :grin::gun:

PR #513, opened 2019-08-01
Show/hide the Runes element itself during Enable/Disable

nihilisticpandemonium:
This will help layouts that have things such as backdrops.

Noticed in ElvUI using NameOnly nameplates would cause the Runes backdrop to still appear because the object iself was not hidden, while the invididual runes were.

### DIFF
--- a/elements/runes.lua
+++ b/elements/runes.lua
@@ -192,6 +192,8 @@ local function Enable(self, unit)
 			end
 		end
 
+		element:Show()
+
 		self:RegisterEvent('PLAYER_SPECIALIZATION_CHANGED', Path)
 		self:RegisterEvent('RUNE_POWER_UPDATE', Path, true)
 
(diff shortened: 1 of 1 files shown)


p3lim:
The element is not intended to be created as a frame, rather a table.
So in cases where the docs are followed, this will fail. You'll have to check if the element is a frame before you try to show/hide it.

nihilisticpandemonium:
Added check for IsObjectType("Frame") around both Show and Hide per your comment.

Rainrider:
I'm not sure I understand what we are fixing here, but checking if Show/Hide methods are defined on the element is more extensible than checking for a single widget type.

nihilisticpandemonium:
[https://i.imgur.com/5NxNWBq.png](url) This is an example of the problem I'm trying to fix.  I did update it to just check for Show and Hide.

Rainrider:
Why can't this be parented to the first rune instead?

Also, when in a vehicle, this border will be still shown and the runes will be hidden.

nihilisticpandemonium:
Closing this as it is an ElvUI issue and we will resolve it in ElvUI.

PR #514, opened 2019-08-19
tags: Escape magic chars in tag names

ls-:
If tag's name contains any of magic chars, there's a chance that `string.match` will fail to find the match.


ls-:
This PR was rendered obsolete by #536.

PR #517, opened 2019-08-26
Remove UNIT_HEALTH_FREQUENT event

ls-:
This PR will remove the UNIT_HEALTH event.

The main issue w/ this event is that it doesn't fire for the last update from whatever percentage to 100%, so, let's say, a health bar will stay at 99% and won't update to 100%. It also cannot be used to update nameplates' health bars reliably.

This event has been broken for the entirety of Legion, in BfA it worked only for a couple of months between 8.0.1 and 8.1.0.


elements/health.lua
```diff
@@ -189,38 +186,16 @@ local function ForceUpdate(element)
...
-		if(element.frequentUpdates) then
-			element.__owner:UnregisterEvent('UNIT_HEALTH', Path)
-			element.__owner:RegisterEvent('UNIT_HEALTH_FREQUENT', Path)
-		else
-			element.__owner:UnregisterEvent('UNIT_HEALTH_FREQUENT', Path)
-			element.__owner:RegisterEvent('UNIT_HEALTH', Path)
-		end
-	end
-end
+local function noop() end
```
p3lim: `nop` is a global function provided by blizz.

elements/health.lua
```diff
@@ -189,38 +186,16 @@ local function ForceUpdate(element)
...
 		element.__owner = self
 		element.ForceUpdate = ForceUpdate
-		element.SetFrequentUpdates = SetFrequentUpdates
-
-		if(element.frequentUpdates) then
-			self:RegisterEvent('UNIT_HEALTH_FREQUENT', Path)
-		else
-			self:RegisterEvent('UNIT_HEALTH', Path)
-		end
+		element.SetFrequentUpdates = noop -- Deprecated
```
p3lim: Remove this, the PR requires a major anyways

elements/health.lua
ls-: orly? O_o

p3lim: https://www.townlong-yak.com/framexml/live/UIParent.lua#5188

ls-: 🤔 👌

PR #519, opened 2019-10-02
castbar: Add fallback icon

Wetxius:
In Legion and BfA some world interaction doen't have icon for casting spell.
You can test it with https://www.wowhead.com/spell=182547/grappling-gun in Stormheim.

### DIFF
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -130,7 +130,7 @@ local function UNIT_SPELLCAST_START(self, event, unit)
 	element:SetValue(0)
 
 	if(element.Text) then element.Text:SetText(text) end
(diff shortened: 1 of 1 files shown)


siweia:
The texture id is 136243.

Rainrider:
Could you please make the fallback texture a constant (just an all caps local) at the top of the file and reference it from the code instead of hard-coding the texture path? A good name would be `CASTBAR_FALLBACK_ICON`. I would also prefer using the texture id instead of the texture path. We can look it up like htts://www.wowhead.com/icon=136243 if the need arises.

Wetxius:
Something like this?

Issue #39, opened 2010-09-05
UnitGetIncomingHeals and UNIT_HEAL_PREDICTION


tekkub:
Yup, incoming heals are int eh API now.  Lets kill HealComm!

PR #521, opened 2019-10-15
Support for Classic

Rainrider:
Not sure about toc fields like interface. Haven't tested much yet.


elements/raidroleindicator.lua
```diff
@@ -42,7 +42,8 @@ local function Update(self, event)
 	end
 
 	local role, isShown
-	if(UnitInRaid(unit) and not UnitHasVehicleUI(unit)) then
+	-- UnitHasVehicleUI is not available on Classic
+	if(UnitInRaid(unit) and UnitHasVehicleUI and not UnitHasVehicleUI(unit)) then
```
p3lim: Use `isRetail` or `isClassic` to signify this instead.

elements/totems.lua
```diff
@@ -47,9 +47,34 @@ OnEnter and OnLeave script handlers will be set to display a Tooltip if the `Tot
...
+		local timeLeft = self.expiration - GetTime()
+		if(timeLeft <= 0) then return end
+
+		local fmt, value = SecondsToTimeAbbrev(timeLeft)
+
+		local time
+		if(fmt == MINUTE_ONELETTER_ABBR) then
+			time = SPELL_TIME_REMAINING_MIN:format(value)
+		else
+			time = SPELL_TIME_REMAINING_SEC:format(timeLeft)
```
p3lim: Consts should have a fallback.

init.lua
```diff
@@ -1,3 +1,6 @@
 local parent, ns = ...
 ns.oUF = {}
 ns.oUF.Private = {}
+
+ns.oUF.isClassic = select(4, GetBuildInfo()) == 11302
+ns.oUF.isRetail = not ns.oUF.isClassic
```
p3lim: I'd stick with just one or the other.

ouf.lua
```diff
@@ -54,7 +55,7 @@ local function updateActiveUnit(self, event, unit)
 		realUnit = 'target'
 	end
 
-	if(modUnit == 'pet' and realUnit ~= 'pet') then
+	if(modUnit == 'pet' and realUnit ~= 'pet' and isRetail) then
```
p3lim: Bool checks are cheaper, do that first.

ouf.lua
```diff
@@ -280,7 +281,7 @@ local function initObject(unit, style, styleFunc, header, ...)
 			objectUnit = objectUnit .. suffix
 		end
 
-		if(not (suffix == 'target' or objectUnit and objectUnit:match('target'))) then
+		if(not (suffix == 'target' or objectUnit and objectUnit:match('target')) and isRetail) then
```
p3lim: Bool checks are cheaper, do that first.

ouf.lua
```diff
@@ -298,7 +299,7 @@ local function initObject(unit, style, styleFunc, header, ...)
 			object:SetAttribute('*type2', 'togglemenu')
 
 			-- No need to enable this for *target frames.
-			if(not (unit:match('target') or suffix == 'target')) then
+			if(not (unit:match('target') or suffix == 'target') and isRetail) then
```
p3lim: Bool checks are cheaper, do that first.

units.lua
```diff
@@ -169,12 +170,12 @@ function oUF:HandleUnit(object, unit)
...
 		object:RegisterEvent('UPDATE_MOUSEOVER_UNIT', object.UpdateAllElements, true)
-	elseif(unit == 'focus') then
+	elseif(unit == 'focus' and not isClassic) then
 		object:RegisterEvent('PLAYER_FOCUS_CHANGED', object.UpdateAllElements, true)
-	elseif(unit:match('boss%d?$')) then
+	elseif(unit:match('boss%d?$') and not isClassic) then
 		object:RegisterEvent('INSTANCE_ENCOUNTER_ENGAGE_UNIT', object.UpdateAllElements, true)
 		object:RegisterEvent('UNIT_TARGETABLE_CHANGED', object.UpdateAllElements)
-	elseif(unit:match('arena%d?$')) then
+	elseif(unit:match('arena%d?$') and not isClassic) then
```
p3lim: Bool checks are cheaper, do that first.

colors.lua
```diff
@@ -116,6 +116,11 @@ colors.power[16] = colors.power.ARCANE_CHARGES
 colors.power[17] = colors.power.FURY
 colors.power[18] = colors.power.PAIN
 
+if(oUF.isClassic) then
+	colors.power[4] = colors.power.HAPPINESS
+	colors.power[14] = colors.power.COMBO_POINTS
+end
```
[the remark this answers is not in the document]
Rainrider: They are always included. Classic just adds 14 for combo points and remaps 4 to happiness.

elements/totems.lua
Rainrider: Should I fallback to the English locale?

init.lua
Rainrider: It is more conveniece than anything else. Had `isClassic` only first then got annoyed having to type `not isClassic`.

colors.lua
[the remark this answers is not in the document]
Rainrider: Btw the classic client has then in as well if this is what you meant.

Wutname1:
What's the reason behind a new element for combopoints when ClassPower is there and can do the job fairly easily. from [wutname1/oUF](https://github.com/Wutname1/oUF/blob/master/elements/classpower.lua) Diff:  https://www.diffchecker.com/RS7anjNR I would think not having 2 different combo point systems makes it so layouts and oUF have less to maintain.

There are a few formatting differences but that's just due to me using VS code format on save.

Wutname1:
You are disabling AdditionalPower this is still used by druids. A functioning version can be viewed here: https://github.com/Wutname1/oUF/blob/master/elements/additionalpower.lua this file has already been updated with #490

Rainrider:
> What's the reason behind a new element for combopoints when ClassPower is there and can do the job fairly easily.

The reason is to strip away unnecessary code. Also, ClassPower is likely to change on retail and I would dislike to have to move those if checks for classic around only to keep the same element on both versions of the game.

> You are disabling AdditionalPower this is still used by druids.

It is actually not. There is no such UI element in Classic. On Retail it is by default only enabled as a pair with the lunar power and only because there are spells castable in a shapeshift form which cost mana rather the power type associated with that form. None of this holds true on Classic, please correct me if I'm wrong. This is just my justification though, I don't mind enabling the element for Classic as well.

Wutname1:
It is used to display a druids mana when shapeshifted. It could be split off I to its own thing as well as that is the only use case I know of. That whole file can trace its roots back to being originally meant for just druid mana.

siweia:
In retail, additionalpower works for shadow priest and elemental shaman as well.

Talyrius:
And Enhancement Shaman.

Rainrider:
Nothing changed in this regard. Or would I break something there with this here?

Talyrius:
I don't believe so. That was in response to this:
> On Retail it is by default only enabled as a pair with the lunar power and only because there are spells castable in a shapeshift form which cost mana rather the power type associated with that form.

The other class/spec use cases for it on retail weren't referenced.

cannonpalms:
Now that incoming heals is no longer provided by the wow api, is undoing the 2010 work to strip HealComm from oUF on the table?

Rainrider:
I don't think oUF had HealComm support in its core package. If you mean something like https://www.wowinterface.com/downloads/info9782-oUF_HealComm.html, the answer is I don't know. My personal view is that it should be done through a plugin like back in the day.

cannonpalms:
Here's the issue I'm referring to. #39 It looks like you are right, HealComm was never a part of oUF core. 

Providing support would ultimately be a matter of implementing the 4 events from HealComm and updating frames appropriately. Nothing terribly difficult.

Rainrider:
Is there an updated LibHealComm for Classic?

Wutname1:
Yes Azilroka has been updating it in elvui and has recently started pushing the code to the old wowace listing  https://www.wowace.com/projects/libhealcomm-4-0/files

oUF.xml
```diff
@@ -44,7 +44,7 @@
 	<!--@end-retail@-->
 
 	<!--@non-retail@-->
-	<Script file="elements\combopoints.lua"/>
+	<Script file="elements\classpowerclassic.lua"/>
```
p3lim: If this is the classic branch, why not just name it "classpower"?

Rainrider: Both files are present in the development version.

The idea is to use the branch to see how long we (I) can reasonably easy keep API parity between retail and classic (and other old expansions if they release it) since we had the discussion internally and this is the main concern and the reason to go with separate branches. Also the branch can be used to create an embedded oUF version supporting both retail and classic for as long as we manage to keep this parity.

Rainrider:
https://www.youtube.com/watch?v=eKmRkS1os7k

Rainrider:
With Blizzard considering a classic BC release, and the possibility it might happen, we don't see a clean and easy way of supporting multiple versions of the game from the same code base and we don't have the bandwidth for it either.

PR #522, opened 2019-11-11
core: add a combat log event system

Rainrider:
Will probably not see much use, but won't hurt either.

### DIFF
--- a/combatevents.lua
+++ b/combatevents.lua
@@ -1,0 +1,116 @@
+local _, ns = ...
+local oUF = ns.oUF
+local Private = oUF.Private
+
+local argcheck = Private.argcheck
+local frame_metatable = Private.frame_metatable
+
+local CombatLogGetCurrentEventInfo = _G.CombatLogGetCurrentEventInfo
+
+local event_metatable = {
+	__call = function(funcs, ...)
+		for self, func in next, funcs do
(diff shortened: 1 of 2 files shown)


p3lim (approved):
Only comment I have on this is that if layouts start mis-using this it'll show blame on oUF for being resource-hungry. A _very_ minor remark, the PR looks good otherwise, but I haven't tested it myself.

ls-:
I like the idea as well, I'm going to test this PR later this week, have my hands full atm. I have a floating combat text plug-in that uses CLEU, I'll try rewriting with this PR in place to see what's missing and what I'd like to be added/removed/changed.

combatevents.lua
```diff
@@ -0,0 +1,116 @@
...
+		for _, func in next, funcs do
+			func(self, ...)
+		end
+	end
+}
+
+local listener = CreateFrame('Frame')
+listener.activeEvents = 0
+
+local function filter(_, event, ...)
```
p3lim, line 31: I'd expand on the name of this function, processEvent perhaps?

Rainrider: It actually filters the events and handles the processing to the registered listener. Event is here the combat log event, maybe I should rather rename this?

PR #528, opened 2020-05-31
save name of target of currently cast spell

LybrialsGit:
Saves the target of the currently cast spell. If so you can do something like that:

```lua
function UNITS:CastBar_PostCastStart(unit)
    local db = self.__owner.db;

    if (db and db.castBar and db.castBar.enabled) then
        if (db.castBar.general.showTarget and self.currentTarget) then
            self.Text:SetText(GetSpellInfo(self.spellID) .. ' > ' .. self.currentTarget);
        end
    end
end
```

example

### DIFF
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -32,6 +32,7 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
 .channeling       - indicates whether the current spell is a channeled cast (boolean)
 .notInterruptible - indicates whether the current spell is interruptible (boolean)
 .spellID          - the spell identifier of the currently cast/channeled spell (number)
+.currentTarget    - the name of the target of the currently cast spell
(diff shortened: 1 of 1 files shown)


p3lim:
This is something that can really easily be done from the layout as needed.

LybrialsGit:
Could you tell me how?

p3lim:
Just register the event in your layout? Something like this:

```lua
self.Castbar:RegisterEvent('UNIT_SPELLCAST_SENT', function(self, event, unit, target)
    self.Castbar.currentTarget = ((target and target ~= '') and target) or nil
end)
```

LybrialsGit:
Oh ok. That is what you meant. From Software Engineering perspective this does not feel right. 
I will use the fork then.

Thanks anyway.

PR #529, opened 2020-08-01
PhaseIndicator: Add Tooltip Support

Rainrider:
This adds tooltip support for the phase indicator for Shadowlands. It is backwards compatible.

### DIFF
--- a/elements/phaseindicator.lua
+++ b/elements/phaseindicator.lua
@@ -7,16 +7,27 @@ Toggles the visibility of an indicator based on the unit's phasing relative to t
 
 PhaseIndicator - Any UI widget.
 
+## Sub-Widgets
+
+Icon - A `Texture` to represent the phased status.
+
 ## Notes
 
 A default texture will be applied if the widget is a Texture and doesn't have a texture or a color set.
+OnEnter and OnLeave script handlers will be set to display a Tooltip if the widget is mouse enabled and does not have
+OnEnter and/or OnLeave handlers.
 
 ## Examples
 
     -- Position and size
-    local PhaseIndicator = self:CreateTexture(nil, 'OVERLAY')
+    local PhaseIndicator = CreateTexture('Frame', nil, self)
     PhaseIndicator:SetSize(16, 16)
     PhaseIndicator:SetPoint('TOPLEFT', self)
+    PhaseIndicator:EnableMouse(true)
+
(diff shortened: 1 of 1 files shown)


elements/phaseindicator.lua
```diff
@@ -39,7 +75,7 @@ local function Update(self, event, unit)
 		element:PreUpdate()
 	end
 
-	local isInSamePhase = UnitInPhase(unit) and not UnitIsWarModePhased(unit)
+	local isInSamePhase = not UnitPhaseReason(unit)
```
p3lim, line 78: I assume this is new in 9.0, and the name of the function doesn't seem to correlate with the return name, I'd expect that to return something else than a boolean. Am I correct in this assumption?

elements/phaseindicator.lua
```diff
@@ -7,16 +7,27 @@ Toggles the visibility of an indicator based on the unit's phasing relative to t
...
 
+## Sub-Widgets
+
+Icon - A `Texture` to represent the phased status.
+
 ## Notes
 
 A default texture will be applied if the widget is a Texture and doesn't have a texture or a color set.
+OnEnter and OnLeave script handlers will be set to display a Tooltip if the widget is mouse enabled and does not have
+OnEnter and/or OnLeave handlers.
```
p3lim, line 18: I'd leave a note about the optional frame vs texture, specifically as the first note here.

Rainrider: I would leave it like it is so we don't have to support it later. The default case is Blizzard like or better - so tooltip it is. The backwards compat is more of a side effect than anything else.

elements/phaseindicator.lua
Rainrider: Yes it returns a ["PhaseReason"](https://github.com/Gethe/wow-ui-source/blob/1ceff078c8866706d2134e5329e1043523f10f73/AddOns/Blizzard_APIDocumentation/UnitDocumentation.lua#L1419-L1432) or nil.

p3lim: Still messes me up that 0 evaluates to true in Lua, even after all these years.

Rainrider:
@p3lim is there anything left for me to address here or could we merge this into the shadowlands branch?

p3lim (approved):

PR #531, opened 2020-08-01
Shadowlands development branch

p3lim:


p3lim:
@Rainrider, @ls-:
Aside from #490 and #522 hopefully merging into this PR, this seems pretty much complete.

We'll take a look at #540 after 10.0.0 has released, if we can't figure out what to do with it before then, as it shouldn't require any breaking changes.

We'll also have to take a look at the new UIWidgets system that is sometimes found on nameplates and evaluate how we should support it in oUF, if at all. I will be taking a look at this some time this week.

Rainrider:
@p3lim we are done with Shadowlands.

Only #522 (no BC break, can be merged later) and #532 (infrastructure, can be merged later) are open but they are both not part of the `shadowlands` milestone.

We shall do a short migration guide. Not sure if we want to merge this into master now though.

elements/additionalpower.lua
```diff
@@ -185,33 +235,50 @@ local function VisibilityPath(self, ...)
...
-	return (self.AdditionalPower.OverrideVisibility or Visibility) (self, ...)
+	(self.AdditionalPower.OverrideVisibility or Visibility) (self, ...)
 end
 
 local function ForceUpdate(element)
-	return VisibilityPath(element.__owner, 'ForceUpdate', element.__owner.unit)
+	VisibilityPath(element.__owner, 'ForceUpdate', element.__owner.unit)
+end
+
+--[[ Power:SetFrequentUpdates(state)
```
p3lim, line 245: missing the `isForced` flag

elements/additionalpower.lua
```diff
@@ -170,10 +202,28 @@ local function Visibility(self, event, unit)
...
+
+		--[[ Callback: AdditionalPower:PostVisibility(isVisible)
+		Called after the element's visibility has been changed.
+
+		* self      - the AdditionalPower element
+		* isVisible - the current visibility state of the element (boolean)
+		--]]
+		if(element.PostVisibility) then
+			element:PostVisibility(true)
+		end
```
p3lim, line 218: This block ends with never calling Path, so the initial update when the element is shown never gets an update.

elements/additionalpower.lua
```diff
@@ -86,44 +90,50 @@ local function UpdateColor(element, cur, max)
...
-
-	* self - the AdditionalPower element
-	* cur  - the current value of the player's additional power (number)
-	* max  - the maximum value of the player's additional power (number)
-	--]]
-	element:UpdateColor(cur, max)
+	element.cur = cur
+	element.max = max
 
 	--[[ Callback: AdditionalPower:PostUpdate(unit, cur, max)
```
p3lim, line 128: Signature is not updated, `unit` should be removed

elements/additionalpower.lua
Rainrider: `ElementEnable` is called in that block and it calls `Path`, which triggers the update.

Rainrider (approved):

Issue #533, opened 2020-08-09
classpower: BugSack error when max power changes


chiel:
Hello, I think I have found an issue with the `classpower` element. When the `max` amount of charges switches for a monk's chi (done through the level 45 "Ascension" talent), the error below gets thrown.

It looks like there may be an issue with the `element.__max` somehow not getting set. Can someone confirm if this is working or if I am just messing stuff up?

I'm creating the element like this:

```lua
	local classPowerFrame = CreateFrame('Frame', nil, UIParent)

	local ClassPower = {}
	for i = 1, 10 do

		local Bar = CreateFrame('StatusBar', nil, self)
		Bar:SetStatusBarTexture(addon.media.texture)
		Bar:SetParent(classPowerFrame)
		Bar:SetSize(16, 6)
		Bar:SetPoint('TOPLEFT', classPowerFrame, 'TOPLEFT', (i - 1) * 17, 0)

		ClassPower[i] = Bar
	end

	self.ClassPower = ClassPower
```

and the error:

```
133x oUF\elements\classpower.lua:129: attempt to compare number with nil
[string "@oUF\elements\classpower.lua"]:129: in function <oUF\elements\classpower.lua:84>
[string "=(tail call)"]: ?
[string "@oUF\events.lua"]:81: in function <oUF\events.lua:79>
[string "=(tail call)"]: ?

Locals:
self = oUF_AelUIplayerPlayer {
 0 = <userdata>
 __tags = <table> {
 }
 UNIT_ENTERED_VEHICLE = <function> defined @oUF\ouf.lua:46
 UNIT_SPELLCAST_CHANNEL_START = <function> defined @oUF\elements\castbar.lua:314
 unitEvents = <table> {
 }
 UNIT_HEALTH_FREQUENT = <function> defined @oUF\elements\health.lua:177
 UNIT_MAXHEALTH = <function> defined @oUF\elements\health.lua:177
 __elements = <table> {
 }
 UNIT_POWER_FREQUENT = <table> {
 }
 Power = <unnamed> {
 }
 UNIT_DISPLAYPOWER = <function> defined @oUF\elements\power.lua:247
 UNIT_SPELLCAST_NOT_INTERRUPTIBLE = <function> defined @oUF\elements\castbar.lua:243
 style = "AelUI:player"
 UNIT_EXITED_VEHICLE = <function> defined @oUF\ouf.lua:46
 UNIT_CONNECTION = <table> {
 }
[...]

PR #534, opened 2020-08-11
classpower: expose the visibility state to layouts through the PostUpdate hook

Rainrider:
- `.isEnabled` is internal state and should not be fiddled with (just like `.__max`), so prepend __.
- layouts depend on this to be able to toggle the visibility of related decorations (see https://github.com/oUF-wow/oUF/issues/533#issuecomment-671592827)

### DIFF
--- a/elements/classpower.lua
+++ b/elements/classpower.lua
@@ -144,9 +144,10 @@ local function Update(self, event, unit, powerType)
 	* max           - the maximum amount of power (number)
 	* hasMaxChanged - indicates whether the maximum amount has changed since the last update (boolean)
 	* powerType     - the active power type (string)
+	* isVisible     - whether the element is visible (boolean?)
 	--]]
 	if(element.PostUpdate) then
-		return element:PostUpdate(cur, max, oldMax ~= max, powerType)
+		return element:PostUpdate(cur, max, oldMax ~= max, powerType, element.__isEnabled)
 	end
 end
 
@@ -184,7 +185,7 @@ local function Visibility(self, event, unit)
 		end
 	end
 
-	local isEnabled = element.isEnabled
+	local isEnabled = element.__isEnabled
 	local powerType = unit == 'vehicle' and 'COMBO_POINTS' or ClassPowerType
 
 	if(shouldEnable) then
@@ -226,7 +227,7 @@ do
 		self:RegisterEvent('UNIT_POWER_FREQUENT', Path)
 		self:RegisterEvent('UNIT_MAXPOWER', Path)
 
-		self.ClassPower.isEnabled = true
+		self.ClassPower.__isEnabled = true
 
 		if(UnitHasVehicleUI('player')) then
 			Path(self, 'ClassPowerEnable', 'vehicle', 'COMBO_POINTS')
@@ -244,7 +245,7 @@ do
 			element[i]:Hide()
 		end
 
-		self.ClassPower.isEnabled = false
+		element.__isEnabled = false
 		Path(self, 'ClassPowerDisable', 'player', ClassPowerType)
 	end
 


elements/classpower.lua
```diff
@@ -144,9 +144,10 @@ local function Update(self, event, unit, powerType)
 	* max           - the maximum amount of power (number)
 	* hasMaxChanged - indicates whether the maximum amount has changed since the last update (boolean)
 	* powerType     - the active power type (string)
+	* isVisible     - whether the element is visible (boolean?)
```
chiel, line 147: Out of curiosity, don't you just want to call this `isEnabled` to match the internal property?

Rainrider: I'm not sure if it would cause confusion with [:IsElementEnabled](https://github.com/oUF-wow/oUF/blob/13f2f6db9109b5025c5b6c89f2fdb6bbb6dddf8b/ouf.lua#L147-L161) which is something different.

ls-:
While I agree that we should prob add the `__` prefix to it, I don't think we should pass is to `Post*`s.

elements/classpower.lua
p3lim: Then the variable should match imo.

chiel: I mean, it's your codebase of course, I'm just a bystander, so it's all up to you guys - but since this is essentially already a thing, introducing a third variable might cause more confusion rather than reduce it?

Rainrider:
@ls- my take on it from a layout author perspective is:

- if it is called `.isEnabled` then it is an option and I am supposed to change it according to my needs
- if it is called `.__isEnabled` then it is internal by convention, which means I shall not change it and its name and even presence are subject to change without prior notice as it is then an implementation detail and not a public API
- if it is internal (on which we for now all agree) it is not part of public documentation. But when layouts have a valid use case which relies on it, we need to provide an accessor API for it. It being a PostUpdate argument is the easiest way here, because it is future proof - if we remove it because implementation changes, we could just pass a Boolean literal to PostUpdate to keep backwards compat until we have an opportunity for a major release.

Is there any argument against it being a PostUpdate argument besides personal preference?

elements/classpower.lua
Rainrider: Yes `.__isVisible` is a better name I suppose

chiel: Hmm, isn't `.isEnabled` used in other elements as well? I guess if you want to change it, it would be good to update it across the board?

Rainrider: [No, it isn't](https://github.com/oUF-wow/oUF/search?q=isEnabled&unscoped_q=isEnabled)

ls-:
@Rainrider, as I said, I'm for the name change, however, even though I do use it in my layout because of all the artwork, so I need to toggle and adjust all the extra bits it doesn't need to be in a `PostUpdate`. There's a lot of undocumented things in oUF that are for internal use only, but can be used by layout devs who like hacking in out our systems. But if you really won't to pass it somehow and somewhere, as a layout dev, I'd like to see separate `Post*`s in our `ClassPowerEnable` and `ClassPowerDisable` that would have the new state of the element. It'd allow to split sometimes annoyingly big `PostUpdate`s into smaller functions that would do layout manipulations in `PostClassPowerEnable|Disable` and just some basic update-related stuff in `PostUpdate`. The element's state change is a fairly rare thing, it doesn't need to be a part of the regular update process, that's why we moved it out of it in the first place.

elements/classpower.lua
chiel: Fair enough! My bad. :)

Rainrider:
@ls- fair enough, I'll go for some PostEnable hooks then. I suppose we would do it for Runes as well, will take a look later or tomorrow.

Rainrider:
@ls- is it good as it is?

PR #537, opened 2020-09-06
alternativepower: replace deprecated UnitAlternatePowerInfo

Rainrider:
This is not backwards compatible because layouts might have relied on `.barType`, `.powerName` and `.powerTooltip` which were previously exposed on the element. `.__barID` and `.__barInfo` are now exposed on the element as a replacement.


Rainrider:
You could test this at [this guy](https://www.wowhead.com/npc=49687/brazie-the-botanist) btw. The quests are available on the premades on the beta realms and he has a daily in case you already completed the series. The code should work on retail as well.

ls- (approved):

PR #538, opened 2020-09-06
classpower: make holy power available in all specs

Rainrider:


p3lim (approved):

p3lim:
I assume all 3 specs have this now in Shadowlands?

Rainrider:
Yes, all three specs have Crusader Strike from level 1. Also [this](https://github.com/Gethe/wow-ui-source/blob/fda9b1b9b1bfd08138522e9fc87d5d5911bad19d/FrameXML/PaladinPowerBar.lua#L10).

PR #539, opened 2020-09-08
runes: fix for death knights without specialization in Shadowlands

Rainrider:
Death Knights start at level 8 without a spec and GetSpecialization() returns 5. Backwards compatible.

### DIFF
--- a/elements/runes.lua
+++ b/elements/runes.lua
@@ -84,7 +84,7 @@ local function UpdateColor(element, runeID)
 	local spec = GetSpecialization() or 0
 
 	local color
-	if(spec ~= 0 and element.colorSpec) then
+	if(spec > 0 and spec < 4 and element.colorSpec) then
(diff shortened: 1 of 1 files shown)


p3lim (approved):

p3lim:
That is just straight up _weird_.

Rainrider:
It appears that as long as there is no spec chosen yet, `GetSpecialization()` returns 5. `GetSpecializationInfo(5)` returns something like that:
```
[1]=1449,
[2]="Initial",
[3]="",
[4]=135846,
[5]="DAMAGER",
[6]=4
```
The first return is a globally unique spec id, so it appears Blizzard handles the base abilities, which now have their own tab in the spellbook, as a separate class spec. `GetNumSpecializations()` is not changed though. Maybe they chose 5 instead of 0 to not break addons' code where 0 is kind of a fallback case (like in our code btw).

PR #540, opened 2020-09-20
classpower: allow layouts to handle charged combo points

Rainrider:
This registers UNIT_POWER_POINT_CHARGE for rogues if the layout provides `ClassPower.UpdateCharged`. This is to support the mechanics of [Echoing Reprimand](https://shadowlands.wowhead.com/spell=323547/echoing-reprimand). [This is how it is handled by the default ui](https://github.com/Gethe/wow-ui-source/blob/6b74238b2eb64bf1ead43eb6208501411b218344/FrameXML/ComboFramePlayer.lua#L218-L241). Because of that we can't really do it in the element and have to push it on layouts.

Please tell if you have a better idea.

### DIFF
--- a/elements/classpower.lua
+++ b/elements/classpower.lua
@@ -302,6 +302,18 @@ local function Enable(self, unit)
 			self:RegisterEvent('UNIT_DISPLAYPOWER', VisibilityPath)
 		end
 
+		--[[ Callback: ClassPower.UpdateCharged(self, event, unit)
+		Used to update charged combo points.
+		While optional, this is solely in the responsibility of the layout.
+
+		* self  - the parent object
+		* event - the event triggering the update (string)
+		* unit  - the unit accompanying the event (string)
+		--]]
+		if(element.UpdateCharged and PlayerClass == 'ROGUE') then
+			self:RegisterEvent('UNIT_POWER_POINT_CHARGE', element.UpdateCharged)
+		end
+
 		element.ClassPowerEnable = ClassPowerEnable
 		element.ClassPowerDisable = ClassPowerDisable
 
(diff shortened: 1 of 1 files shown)


Rainrider:
This is lacking. I'll close it for now and figure something out later.

Rainrider:
This is a rather simple way to handle it: register UNIT_POWER_POINT_CHARGE and let layouts handle it in PostUpdate, more or less like the stock UI implementation.

The problems with the previous solution were:
- it wasn't updated on PEW (so it either had to go through the normal update `Path` or the `UpdateCharged` check had to be added to `Path` as well).
- depending on their use-case layouts may need the current and the max amount of combo points (for color and position/size), so using `Path` as the event handler seems the way to go.

Downside of the current implementation:
- it is not immediately apparent how layouts are supposed to handle the charged combo points. We might mention this in the migration guide though.

ls-:
Potentially we could register `Path` for `UNIT_POWER_POINT_CHARGE` for rogues and droids, and then we'd pass, let's say, `chardgedIdx` as an extra arg to `PostUpdate`. From blizz code
```lua
local chargedPowerPoints = GetUnitChargedPowerPoints(self:GetParent().unit);
-- there's only going to be 1 max
local chargedPowerPointIndex = chargedPowerPoints and chargedPowerPoints[1];
```
In our case  `chargedPowerPointIndex ` will become `chardgedIdx`.

It doesn't fire that often to begin with, so it should be a major problem in terms of performance.

Rainrider:
@ls- we have to update it on every update for rogues at least. As far as I checked, only Assassination Rogues could have Echoing Reprimand. Haven't found anything else that uses this mechanic. I could trottle it by using the `powerType` though (like only for combos) to play it safe if you think that's better.

ls-:
@Rainrider ok, since it's tied to `Echoing Reprimand` exclusively, we could throttle it even further by adding a similar case, but for rogues https://github.com/oUF-wow/oUF/blob/c143a39ef22d24f354a46e63c258d147456a23b8/elements/classpower.lua#L291-L294.

ls-:
@Rainrider yeah, since it's tied to `Echoing Reprimand` exclusively, limiting it rogues only makes sense, I thought droids could use it too since Blizz don't check for class in any way and I don't play dudu. My b. So far, the only issues I have is this var name `chargedPowerPointIndex`, obnoxiously long, `chargedIndex` should suffice, or given that it's fetched from the `chargedPoints` table, it could also be `chargedPoint`, other than that LGTM.

ls- (approved):
LGTM

PR #544, opened 2020-10-10
Tags: Fix suffix delimiter docs

gizzmo:
It's also in listed in the 10.0 update thread as `<$`. IMO that does look better.


ls-:
Not that you mention it, `$>` and `<$` indeed look cuter.

Issue #546, opened 2020-10-10
oUF.colors.power usage inconsistencies


myzb:
With the new color path changes most of the elements that use the  ``oUF.colors.power`` table retrieve the color information by passing the color index (aka 'powerType' ID) i.e:

https://github.com/oUF-wow/oUF/blob/master/elements/alternativepower.lua#L89
https://github.com/oUF-wow/oUF/blob/master/elements/additionalpower.lua#L73

For power however color retrieval is being done using the actual token (MANA, RAGE, etc) first:
``t = self.colors.power[ptoken or ptype]``

See:
https://github.com/oUF-wow/oUF/commit/2ca03b77f0fefcf03501bb176b410ae8f698af6a#diff-2d6dd4967e1f58a5832a8c6702fa31c9R125

This means that to i.e override the mana color 2 declarations are now needed:
```
oUF.colors.power[0] = { 1/255, 121/255, 228/255 }       -- for alternativepower and additionalpower
oUF.colors.power['MANA'] = { 1/255, 121/255, 228/255 }  -- for power
```

Is this done on purpose? It seems we want the following with ``ptype`` first instead?
```
t = self.colors.power[ptype or ptoken]
```

Also, documentation would need to get adjusted
https://github.com/oUF-wow/oUF/blob/master/elements/power.lua#L35

OR:

perhaps use ``ADDITIONAL_POWER_BAR_NAME`` instead of ``ADDITIONAL_POWER_BAR_INDEX`` for color retrieval in alternativepower and additionalpower?

ls-:
Sup!
```lua
t = self.colors.power[ptoken or ptype]
```
It's done this way because of [this](https://github.com/Gethe/wow-ui-source/blob/beta/FrameXML/CompactUnitFrame.lua#L509-L522). It's a very old piece of code that's added back in Cata I think and no one really bothered to change it. Technically, it should be
```lua
t = self.colors.power[ptoken] or self.colors.power[ptype]
```
As for
> This means that to i.e override the mana color 2 declarations are now needed

That's always the case, in oUF the default values we use come from [Blizz's PowerBarColor table](https://github.com/Gethe/wow-ui-source/blob/beta/FrameXML/UnitFrame.lua#L2-L39) which contains both types and tokens, so we use the same structure. Also, in this new release [I added alt power to our colour table](https://github.com/oUF-wow/oUF/blob/master/colors.lua#L121-L122).

myzb:
Thanks for the info

[...]

PR #551, opened 2020-10-14
Fix additional power color

siweia:


ls-:
Sup!
Alt power bar fix looks weird, alt power colour can be accessed via both index and name because of [this](https://github.com/oUF-wow/oUF/blob/master/colors.lua#L121-L122).
If you use custom alt power colour, how do you override it? Could you post dumps of your unit frame's colour table?
```
/dump UFNAME.colors.power[Enum.PowerType.Alternate]
/dump UFNAME.colors.power.ALTERNATE
```

siweia:
It is within additional power, not alt power.

ls-:
I'm fooking blind, mate 😁 🔫 But the same questions still stand. Post your dumps
```
/dump UFNAME.colors.power[ADDITIONAL_POWER_BAR_NAME]
/dump UFNAME.colors.power[ADDITIONAL_POWER_BAR_INDEX]
```
Should be something like this:
image

siweia:
I change the MANA color in my own layout.
`oUF.colors.power.MANA = {0, .4, 1}`

Here is the dump:

ls-:
When you override any colour you have to override at its index AND its name. It's also brought up in https://github.com/oUF-wow/oUF/issues/546 recently. I also explained why you have to do it that way.

siweia:
Ok, thanks.

ls-:
You're welcome ❤

PR #552, opened 2020-10-14
Fix additional power color show

EKE00372:


ls-:
I know you closed it, but since you encountered it and created a PR, it means that you don't know why it happened. Please read #551 for details.

PR #557, opened 2020-10-17
Lint

Rainrider:
I got tired of ignoring luacheck, so let's start this :)

Basic ideas:
- prepend `_G.` to global constants and addons stuff to keep `.luacheckrc` smaller
- ~~use `strsplit` instead of `string.split` as I was not able to figure out how to tell luacheck to shut up about it~~
- made an exception about unused self because it makes us switch from colon to dot syntax in quite some places
- ignored unused event because maybe it is ok to let it in for debugging elements
- inline ignored unused private.error and prive.print for obvious reasons
- ignored too long lines because we have too much of those right now and I don't care that much
- ignored shadowing as it is too much work to change now
- maybe I forgot something, not sure

Will add [luacheck](https://github.com/nebularg/actions-luacheck) to the github workflow as well if you like this and don't mind.


.luacheckrc
```diff
@@ -0,0 +1,131 @@
...
+	'631', -- line is too long
+}
+
+read_globals = {
+	'debugstack',
+	'strjoin',
+	'strsplit',
+	'wipe',
+
+	-- category?
```
p3lim, line 19: FrameXML

p3lim:
So I assume the .luacheckrc is populated manually?

p3lim:
We'll probably want to add this as an action to run on PRs, and set that as a merge blocker.

Rainrider:
> So I assume the .luacheckrc is populated manually?

Yes, I think it is better to just add stuff as we go.

> We'll probably want to add this as an action to run on PRs, and set that as a merge blocker, no need to add it to the release workflow.

Yep, makes sense.

FireSiku:
Commenting on this pretty late but in regards to this bullet point:
- use strsplit instead of string.split as I was not able to figure out how to tell Luacheck to shut up about it

Luacheck allows you to define globals with restricted fields, and "split" isn't one of those fields. You can define it by adding it to the read_globals and it won't complain anymore.
```
read_globals = {
	string = { fields = { 'join', 'split', 'trim' }},
	table = { fields = { 'wipe' }},
	'debugstack',
	...
```

elements/auras.lua
```diff
@@ -146,7 +146,7 @@ local function createAuraIcon(element, index)
 	return button
 end
 
-local function customFilter(element, unit, button, name)
+local function customFilter(element, _, button, name)
```
p3lim, line 149: I'm not a fan of this one, since it "hides" the reference implementation signature (although it _is_ documented further below so it's probably fine).

Rainrider: I could ignore unused arguments altogether:
```diff
ignore = {
-	'2/self', -- unused argument self
-	'2/event', -- unused argument event
+	'212', -- unused arguments
	'3/event', -- unused value event
	'4', -- shadowing
	'631', -- line is too long
}
```

Rainrider: Or we could ignore unused arguments for elements only:
```
files["**/elements/*.lua"].ignore = {
	'212', -- unused arguments
}
```

Rainrider: We could also use `--luacheck: no unused` at the end of the line to turn it off just there. Those are all options I think. I find relying on the docs alright though, but I understand what you mean.

ouf.lua
```diff
@@ -689,21 +689,21 @@ do
...
+			SecureHandlerSetFrameRef(header, 'clickcast_header', _G.Clique.header)
 		end
 
 		if(header:GetAttribute('showParty')) then
 			self:DisableBlizzard('party')
 		end
 
 		if(visibility) then
-			local type, list = string.split(' ', visibility, 2)
+			local type, list = strsplit(' ', visibility, 2)
```
p3lim, line 701: Wasn't this whitelisted in .luacheckrc?

ouf.lua
```diff
@@ -689,21 +689,21 @@ do
...
 			if(list and type == 'custom') then
 				RegisterAttributeDriver(header, 'state-visibility', list)
 				header.visibility = list
 			else
-				local condition = getCondition(string.split(',', visibility))
+				local condition = getCondition(strsplit(',', visibility))
```
p3lim, line 706: Wasn't this whitelisted in .luacheckrc?

Rainrider: Both `string.split` and `strsplit` are whitelisted now. The latter is a micro speed optimisation, that's why I didn't bother to change it back.

elements/auras.lua
ls-: I agree with @p3lim on this one. In all honestly, unused arguments aren't really an issue, if there's an unused arg, in most cases it's there for a reason, for instance, like what p3lim described. Alternatively, someone messed while implementing stuff. So, imho, we should highlight unused args with the exception of unused `self`. Also, inlining is kaka, don't do it.

Rainrider: We decided to use `--luacheck: no unused` where it makes sense to keep the documentation value if the argument names.

Rainrider: We decided to use more specific exceptions (212, 312) for arguments named `event` and `unit`.

p3lim (approved):
LGTM, need to try out that workflow tho, but can't really do that without merging first

PR #560, opened 2020-10-21
phaseindicator: Pass phaseReason to PostUpdate

ls-:
In Shadowlands there's now 4 different types of phases, so simply saying that the unit is phased or nah is no longer enough. I kept `isInSamePhase` for the sake of backwards-compatibility, but in the future only `phaseReason` will be passed to `PostUpdate` since if it's `nil` then units are in the same phase.

### DIFF
--- a/elements/phaseindicator.lua
+++ b/elements/phaseindicator.lua
@@ -42,19 +42,20 @@ Used to populate the tooltip when the widget is hovered.
 * self - the PhaseIndicator widget
 --]]
 local function UpdateTooltip(element)
-	local unit = element.__owner.unit
-	local reason = UnitPhaseReason(unit)
-	local text = reason and PartyUtil.GetPhasedReasonString(reason, unit) or ''
-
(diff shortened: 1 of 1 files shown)


Rainrider (approved):
🆗

elements/phaseindicator.lua
```diff
@@ -75,21 +76,24 @@ local function Update(self, event, unit)
 		element:PreUpdate()
 	end
 
-	local isInSamePhase = not UnitPhaseReason(unit)
-	if(not isInSamePhase and UnitIsPlayer(unit) and UnitIsConnected(unit)) then
+	local phaseReason = UnitIsConnected(unit) and UnitPhaseReason(unit) or nil
```
p3lim, line 79: Why the `or nil` at the end here?

ls-: For the sake of consistency, tbhon, that's one of `UnitPhaseReason`'s expected returns.

PR #563, opened 2020-11-06
Event wrapper

ls-:
In early-mid BfA `RegisterEvent`, `RegisterUnitEvent`, and `UnregisterEvent` started throwing errors, it wasn't a problem until recently, in SL Blizz started renaming and removing events which in turn started breaking oUF which wouldn't happen prior to the aforementioned error change, it'd just fail silently and wouldn't affect the framework. Wrapping these functions, where it's needed, with `xpcall`s will give layout devs, plugin devs, and us extra time to gather, report, and/or react to this kind of issues without breaking oUF itself since it will load and continue to function properly despite all the event errors.

This **WILL NOT** stop error messages from popping up.

### DIFF
--- a/elements/tags.lua
+++ b/elements/tags.lua
@@ -106,6 +106,7 @@ local _, ns = ...
 local oUF = ns.oUF
 local Private = oUF.Private
 
+local xpcall = Private.xpcall
 local unitExists = Private.unitExists
 
 local _PATTERN = '%[..-%]+'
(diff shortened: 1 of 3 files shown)


p3lim:
> oUF itself since it will load and continue to function properly

Well, it probably won't function properly, but I see the point.

p3lim (dismissed):
LGTM

Rainrider:
LGTM as well :)

p3lim (approved):

PR #565, opened 2020-11-18
Event wrapper fix

ls-:
As it's brought to my attention by @myzb, #563 kinda broke headers due to the fact that at the very beginning headers might not have legit units, so they don't pass the unit check, so the code doesn't even reach the event registration part. This PR addresses the issue by replacing all the wrappers with one `validateEvent` call.

### DIFF
--- a/events.lua
+++ b/events.lua
@@ -3,8 +3,8 @@ local oUF = ns.oUF
 local Private = oUF.Private
 
 local argcheck = Private.argcheck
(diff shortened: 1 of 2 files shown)


myzb:
The core parts of oUF are still a bit to much magic for me, but I can atleast confirm that this PR fixes the issues with header units.

p3lim (approved):

PR #566, opened 2020-11-22
tags: Untag fontstrings correctly

p3lim:
When calling `table.remove` on an indexed table the whole table is shifted to fill the gap, thus `next` will skip one fontstring and call it a day. This has the potential of leaving the fontstring in the events table and still being processed when updating tags, resulting in an error from the eventhandler trying to call `UpdateTag` on it.


ls- (approved):
👌

Issue #570, opened 2021-01-02
Aura Tooltip makes GameTooltip "leave the screen"


LybrialsGit:
**Describe the bug**
After you hover over a oUF nameplate aura, the GameTooltip will no longer clamp to screen.
It will make the GameTooltip "leave the screen" and so cut off information.

**How to reproduce**
1. Tooltips are not leaving screen:

![tooltip not leaving screen](https://i.imgur.com/43rSfeo.png)

2. I hover an aura on a nameplate to check the aura tooltip

![Hover nameplate aura](https://i.imgur.com/bVutJxi.png)

3. Tooltips are now leaving screen:

![tooltip leaving screen](https://i.imgur.com/iWxodbK.png)

**What did you expect?**
I would expect that the GameTooltip keeps to clamp to the screen.

**Error report**
No Error Report

**Screenshots**
![tooltip not leaving screen](https://i.imgur.com/43rSfeo.png)
![Hover nameplate aura](https://i.imgur.com/bVutJxi.png)
![tooltip leaving screen](https://i.imgur.com/iWxodbK.png)

Rainrider:
This looks quite a lot like what was reported in #510 and was fixed by #511 and #512. Maybe that's not the right way to detect anchoring restrichtions anymore? @ls- ?

LybrialsGit:
Hi, 

I found out that this only happens if I set the `tooltipAnchor` in the auras config manually.
If I do not set this option the error does not occur. 

For example If I do:

```
auras.tooltipAnchor = 'ANCHOR_BOTTOMRIGHT'
```

which is the default If I read that correctly, the error occurs as I reported it. 
But If I do not set this configuration everything works fine.

ls-:
@Rainrider, it's definitely related to notoriously not-so-great Blizz tooltip code, it's working fine previously, but I guess they changed something with SL. I'll test it with `"ANCHOR_CURSOR_RIGHT"` as an alternative, since that's what Blizz use in their UI code now. Alternatively, we could replace `GameTooltip` with our own tooltip and use it in all elements.

p3lim:
Can you try without TSM and Pawn?

ls-:
@p3lim, I already tried that locally since I use TSM all the time, I also installed Pawn for testing since I noticed it in his screenshots. Everything is working fine, at least here. Weird.

LybrialsGit:
I tested it with my nameplate addon and my core addon being the only active addons and the following code for auras:

```lua
[...]

PR #571, opened 2021-01-04
auras: Update workaround for frames with anchoring restrictions

ls-:
This PR addresses: #570. This happens when users update options on the fly which my previous workaround didn't take into consideration, prob because I don't use that option myself 🥺

I also added tests for forbidden `GameTooltip`s, it's not super common, but it happens, I personally encountered it few times.


Rainrider (approved):
LGTM

p3lim:
Fixes #571.

PR #582, opened 2021-04-23
Keep Aspect Ratio

Thatmemedk:
Would be nice to do this way, so you can do use Keep Aspect Ratio on Auras 💯


p3lim:
Add backwards compatibility and I'll accept it.

p3lim:
@Thatmemedk ping

p3lim (approved):

PR #589, opened 2021-08-30
[WIP] Auras revamp

ls-:
Originally, the goal was to add a sorting hook so layouts could sort auras however they needed. It's a nice feature to have, [even Blizz aura headers have a bunch custom sorting options](https://github.com/Gethe/wow-ui-source/blob/live/FrameXML/SecureGroupHeaders.lua#L637-L640), we, however, do not 😥. But it turned out that adding it to our current implementation would be impossible because right now each aura button goes through a complete update process individually.

So far I split the update process into 4 stages:
- filtering;
- sorting;
- redrawing;
- positioning.

It's still pretty rough, I'll try to stream line it a bit more.

⚠ These changes aren't backwards compatible! ⚠


ls-:
Just to give it some closure. For the time being I'm scrapping the idea of revamping auras, I'll get back to it in 11.x or whenever they decide to deprecate `UnitAura`. I tested it again in beta with the new `C_UnitAuras` API and it's just not that good. While it ties everything together really nicely for the most part, we still need to use both `slot`s and `auraInstanceID`s instead of just using `auraInstanceID`s for everything, in terms of performance it's just meh. It's inconsistent, on average it's ~1.5x slower than the current method, but sometimes it gets as bad as 9x due to some weird hiccups/lag. It's most likely due to the fact that all new methods return packed data, like, let's say, `C_UnitAuras.GetAuraDataBySlot` is 4 times slower than `UnitAuraBySlot` on average, the only difference is that the latter is the old method that returns unpacked data. And this stuff adds up.
Atm the only benefit to using revamped `UNIT_AURA` and the new API is the ability to skip updates, sometimes `UA` fires, but there's nothing to update, and you can bail out earlier which does save a lot of resources, HOWEVER, it doesn't happen as often to justify the overall performance hit.

PR #592, opened 2021-08-31
leaderindicator: Differentiate between leaders and guides

ls-:
It's been a while since these were added, but there's actually 2 types of leaders, guides and leaders, the former are leaders of parties formed via LFD/LFR. [In the default](https://github.com/Gethe/wow-ui-source/blob/live/Interface/FrameXML/PartyMemberFrame.lua#L231-L242) [UI they use](https://github.com/Gethe/wow-ui-source/blob/live/Interface/FrameXML/TargetFrame.lua#L131-L142) [different icons](https://github.com/Gethe/wow-ui-source/blob/live/Interface/FrameXML/PlayerFrame.lua#L97-L110). Also, those `UnitIn*` checks are unnecessary.


.luacheckrc
```diff
@@ -69,6 +69,7 @@ read_globals = {
 	'GetUnitPowerBarInfo',
 	'GetUnitPowerBarInfoByID',
 	'GetUnitPowerBarStringsByID',
+	'HasLFGRestrictions',
```
Rainrider, line 72: Is this wrongly indented or is it the damn GitHub mobile app?

Rainrider (commented):
Not sure if it is the rendering of the diff in the app, sorry if so

.luacheckrc
ls-: image
All tabs.

Rainrider (approved):

ls-:
This PR went full circle 😁🔫

Rainrider:
@ls- do you want to remove default textures next expac or why the compat mode thingy?

ls-:
@Rainrider I originally wanted to push this stuff during SL, so I added the compat mode thingy, but I decided to delay it until the next xpack, so I'll remove it later on.

siweia:
Is it possible to merge LeaderIndicator and AssistantIndicator into one? There shouldn't be someone mark as both roles, right?

ls-:
> Is it possible to merge LeaderIndicator and AssistantIndicator into one? There shouldn't be someone mark as both roles, right?

Given the new functionality of the leader indicator, I'm not sure about that. But I'll review it for the next major. We'll see.

PR #599, opened 2022-02-27
core: Check if event is unitless

p3lim:
Throw a custom error whenever something attempts to register a
unitless event without the unitless flag.


ls- (approved):
👌🏼

PR #600, opened 2022-02-27
core: Enable arena frames if layout configures them

p3lim:


p3lim:
Nvm just me screwing something up, we actually specifically disable this in blizzard.lua

PR #603, opened 2022-03-22
core: Rework how eventless units are handled

ls-:
Had to rework the way we handle eventless or "invalid" units because unlike `"boss1..5"` new `"boss6..8"` units can't be used for `RUE`, it silently fails and falls back to `RE`. The backend seems to be f*cking raw™. Along the way cleaned the code up a bit and fixed a couple of bugs I stumbled upon 😋 

So now instead of using internal `Private.enableTargetUpdate(object)` there's `oUF:HandleEventlessUnit(object)` that can be overridden by users in case they need to do something else in there.

Also, instead of creating a new `OnUpdate` for each frame we now handle eventless unit frames the same way we handle eventless tags, so there's one `OnUpdate` per timer.

### DIFF
--- a/ouf.lua
+++ b/ouf.lua
@@ -24,25 +24,6 @@ PetBattleFrameHider:SetAllPoints()
 PetBattleFrameHider:SetFrameStrata('LOW')
 RegisterStateDriver(PetBattleFrameHider, 'visibility', '[petbattle] hide; show')
 
--- updating of "invalid" units.
-local function enableTargetUpdate(object)
-	object.onUpdateFrequency = object.onUpdateFrequency or .5
-	object.__eventless = true
-
-	local total = 0
-	object:SetScript('OnUpdate', function(self, elapsed)
-		if(not self.unit) then
-			return
-		elseif(total > self.onUpdateFrequency) then
-			self:UpdateAllElements('OnUpdate')
-			total = 0
-		end
-
-		total = total + elapsed
-	end)
-end
-Private.enableTargetUpdate = enableTargetUpdate
-
 local function updateActiveUnit(self, event)
 	-- Calculate units to work with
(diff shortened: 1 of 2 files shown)


ouf.lua
```diff
@@ -262,13 +243,31 @@ local function updateRaid(self, event)
 	end
 end
 
+-- boss6-8 exsist in some encounters, but unit event registration seems to be
+-- completely broken for them, so instead we use OnUpdate to update them.
+local eventlessUnits = {
+	['boss6'] = true,
```
Rainrider, line 249: `boss6 = true` would be more consistent with the current code style. See https://github.com/oUF-wow/oUF/blob/master/events.lua#L19-L29. On the other hand we break this in tags.lua, so feel free to ignore this.

ouf.lua
```diff
@@ -308,15 +301,11 @@ local function initObject(unit, style, styleFunc, header, ...)
...
-			end
-
-			-- Other boss and target units are handled by :HandleUnit().
-			if(suffix == 'target') then
-				enableTargetUpdate(object)
+			if(isEventlessUnit(objectUnit)) then
+				oUF:HandleEventlessUnit(object)
 			else
+				-- No need to enable this for eventless units.
+				object:SetAttribute('toggleForVehicle', true)
```
Rainrider, line 308: Does the `toggleForVehicle` attribute depend on events? Is there any harm to keep this enabled for eventless units?

Rainrider (commented):
LGTM so far and it didn't break anything in my layout at least :) Would be nice if you'd comment on `toggleForVehicle`

ouf.lua
ls-: No, not really, it's used in `SecureButton_GetModifiedUnit`, so technically we could enable it for eventless units.

Rainrider (approved):

Rainrider (approved):

PR #604, opened 2022-03-27
core: use ColorMixin for colors

### LINKED ISSUE #590: Use ColorMixin
Consider using [ColorMixin](https://www.townlong-yak.com/framexml/latest/Color.lua) for everything in the [oUF.colors namespace](https://github.com/oUF-wow/oUF/blob/master/colors.lua).
This would let us use the handy methods from the mixin, including `:GetRGB()` instead of unpacking the color object like we do today. Blizzard is using this for a decent amount of their own color objects (e.g. RAID_CLASS_COLORS), and so should we.

This is obviously not possible to make backwards compatible, so I'm putting it up for discussion before next major.

Rainrider:
Closes #590

No BC breaks. I'll switch elements and stuff if you want this. Just left them in for now to demonstrate BC.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -25,6 +25,7 @@ read_globals = {
 	'CastingBarFrame_OnLoad',
 	'CastingBarFrame_SetUnit',
 	'ComboFrame',
+	'CreateColor',
 	'Enum',
 	'FocusFrame',
 	'MonkStaggerBar',
--- a/colors.lua
+++ b/colors.lua
@@ -4,35 +4,44 @@ local Private = oUF.Private
 
 local frame_metatable = Private.frame_metatable
 
+function oUF:CreateColor(r, g, b, a, atlas)
+	local color = CreateColor(r, g, b, a)
+
+	color[1], color[2], color[3] = r, g, b
+	color.atlas = atlas
+
+	return color
+end
+
 local colors = {
 	smooth = {
 		1, 0, 0,
 		1, 1, 0,
 		0, 1, 0
 	},
-	health = {49 / 255, 207 / 255, 37 / 255},
-	disconnected = {0.6, 0.6, 0.6},
-	tapped = {0.6, 0.6, 0.6},
+	health = oUF:CreateColor(49 / 255, 207 / 255, 37 / 255),
+	disconnected = oUF:CreateColor(0.6, 0.6, 0.6),
+	tapped = oUF:CreateColor(0.6, 0.6, 0.6),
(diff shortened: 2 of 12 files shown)


p3lim (approved):
LGTM

p3lim:
If you want to update everything else to use :GetRGB() and whatnot that's up to you, but it'll eventually be something we'll want to do.

This probably also involves updating the color-related APIs oUF exposes, but they can probably be left alone unless there are performance reasons to update them.

Rainrider:
Used following code for a perf benchmark where `color` is a local copy of `Interface/SharedXML/Color.lua` and `Interface/SharedXML/Mixin.lua` (only what is needed for the call to CreateFromMixins):

```lua
package.path = package.path .. ';./?.lua'

require('color')

local function benchmark(times, name, f, ...)
	local start = os.clock()
	for i = 1, times do
		f(...)
	end
	local elapsed = (os.clock() - start)

	print(('%16s: %9.3f ms'):format(name, elapsed * 1e4))
end

local r, g, b = 0.1, 0.4, 0.8
local color = CreateColor(r, g, b)
color[1] = r
color[2] = g
color[3] = b
color.atlas = 'some-atlas'

local color2 = { r, g, b, atlas = 'some-atlas' }

local function callRGB(color)
	local r, g, b = color:GetRGB()
end

local function callUnpack(color)
	local r, g, b = unpack(color)
end

local function getByIndex(color)
	local r, g, b = color[1], color[2], color[3]
end

local function getByName(color)
	local r, g, b = color.r, color.g, color.b
end

benchmark(1e6, 'color:getRGB()', callRGB, color)
benchmark(1e6, 'unpack(mixin)', callUnpack, color)
benchmark(1e6, 'index mixin', getByIndex, color)
benchmark(1e6, 'name mixin', getByName, color)
benchmark(1e6, 'unpack(simple)', callUnpack, color2)
benchmark(1e6, 'index simple', getByIndex, color2)
benchmark(1e6, 'name simple', getByName, color2)
```
Results:
```
  color:getRGB():   795.010 ms
   unpack(mixin):  1178.900 ms
     index mixin:   591.340 ms
      name mixin:   532.440 ms
  unpack(simple):   970.390 ms
    index simple:   689.990 ms
     name simple:   684.070 ms
```

So I kept the direct table access instead of calling `color:GetRGB()` but retrieve the color components by name rather than index. This allows us to ditch the array part in the next expac.

p3lim (approved):
:+1:

PR #607, opened 2022-04-16
ci: check code style

Rainrider:
With the risk you guys might hate me:

- use [StyLua](https://github.com/JohnnyMorganz/StyLua) for code formatting

Pros:

- it is opinionated, so hardly any space for discussion (see [Configuration](https://github.com/JohnnyMorganz/StyLua#configuration) though)
- we can [ignore part of files](https://github.com/JohnnyMorganz/StyLua#ignoring-parts-of-a-file) or [whole files](https://github.com/JohnnyMorganz/StyLua#filtering-using-styluaignore) as well
- it has a GitHub action
- is has plugins for all the editors we use:
  - vscode
  - sublime
  - neovim
- it does not support "our" if syntax (this is the biggest gain for me :))

Cons:

- zero (well this is just my subjective opinion :))

And because I know this might become a discussion point:

```diff
--- a/elements/alternativepower.lua
+++ b/elements/alternativepower.lua
@@ -199,9 +198,13 @@ local function Visibility(self, event, unit)
 	local barInfo = GetUnitPowerBarInfoByID(barID)
 	element.__barID = barID
 	element.__barInfo = barInfo
-	if(barInfo and (barInfo.showOnRaid and (UnitInParty(unit) or UnitInRaid(unit))
-		or not barInfo.hideFromOthers
-		or UnitIsUnit(unit, 'player')))
+	if
+		barInfo
+		and (
+			barInfo.showOnRaid and (UnitInParty(unit) or UnitInRaid(unit))
+			or not barInfo.hideFromOthers
+			or UnitIsUnit(unit, 'player')
+		)
 	then
 		self:RegisterEvent('UNIT_POWER_UPDATE', Path)
 		self:RegisterEvent('UNIT_MAXPOWER', Path)
```

So, what do you think? Btw see the [ci job output](https://github.com/oUF-wow/oUF/runs/6047801164?check_suite_focus=true#step:3:8) for what would be changed if you accept this.

### DIFF
--- a/elements/auras.lua
+++ b/elements/auras.lua
@@ -76,13 +76,17 @@ local VISIBLE = 1
 local HIDDEN = 0
 
 local function UpdateTooltip(self)
-	if(GameTooltip:IsForbidden()) then return end
+	if GameTooltip:IsForbidden() then
+		return
+	end
 
 	GameTooltip:SetUnitAura(self:GetParent().__owner.unit, self:GetID(), self.filter)
 end
 
 local function onEnter(self)
-	if(GameTooltip:IsForbidden() or not self:IsVisible()) then return end
+	if GameTooltip:IsForbidden() or not self:IsVisible() then
+		return
+	end
 
 	-- Avoid parenting GameTooltip to frames with anchoring restrictions,
 	-- otherwise it'll inherit said restrictions which will cause issues with
@@ -92,7 +96,9 @@ local function onEnter(self)
 end
 
 local function onLeave()
-	if(GameTooltip:IsForbidden()) then return end
+	if GameTooltip:IsForbidden() then
+		return
+	end
 
 	GameTooltip:Hide()
 end
@@ -117,7 +123,7 @@ local function createAuraIcon(element, index)
 	local overlay = button:CreateTexture(nil, 'OVERLAY')
 	overlay:SetTexture([[Interface\Buttons\UI-Debuff-Overlays]])
 	overlay:SetAllPoints()
-	overlay:SetTexCoord(.296875, .5703125, 0, .515625)
+	overlay:SetTexCoord(0.296875, 0.5703125, 0, 0.515625)
 	button.overlay = overlay
 
 	local stealable = button:CreateTexture(nil, 'OVERLAY')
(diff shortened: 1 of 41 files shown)


p3lim:
> it does not support "our" if syntax (this is the biggest gain for me :))

I assume you mean the style guide we've defined, and why is this "the biggest gain" for you?

That example from alternativepower is, in my eyes, disgusting :stuck_out_tongue:.

Rainrider:
> I assume you mean the style guide we've defined, and why is this "the biggest gain" for you?

I find the current if-style less readable and unintuitive. There is no tooling to automate this style, so I could write my own and convert to it automatically.

Generally being able to format code automatically is a gain in my eyes. The one "odd" case I found in the diff is maybe an overkill but this is not uncommon in PHP for example so it actually looks ok to me. However, we could use the ignore range for this one if you are so disgusted by it.

p3lim:
Lua is an unopinionated language, and I'm all for something like this, as it resembles gofmt from Go. The alternativepower example was disgusting before as well, but the opinionated styling isn't much better.

Could you run StyLua on the entire codebase and commit it so we can see a full comparison?

p3lim:
Sidenote: we're now getting quite a lot of meta files in the root directory, perhaps we should consider moving the "core" files into core/

Rainrider:
The block ignore markers are broken currently.

Rainrider:
ouf.lua might need some massaging :) Will take a look later

elements/auras.lua
```diff
@@ -194,25 +205,49 @@ local function updateIcon(element, unit, index, offset, filter, isDebuff, visibl
...
+			spellID,
+			canApply,
+			isBossDebuff,
+			casterIsPlayer,
+			nameplateShowAll,
+			timeMod,
+			effect1,
+			effect2,
+			effect3
+		)
```
p3lim, line 230: Great example where splitting arguments into newlines can be a bad idea.

elements/auras.lua
```diff
@@ -194,25 +205,49 @@ local function updateIcon(element, unit, index, offset, filter, isDebuff, visibl
...
 			end
 
-			if(button.overlay) then
-				if((isDebuff and element.showDebuffType) or (not isDebuff and element.showBuffType) or element.showType) then
+			if button.overlay then
+				if
+					(isDebuff and element.showDebuffType)
+					or (not isDebuff and element.showBuffType)
+					or element.showType
+				then
```
p3lim, line 250: This one decreases readability.

elements/auras.lua
```diff
@@ -316,42 +359,64 @@ local function filterIcons(element, unit, filter, limit, isDebuff, offset, dontH
...
-		local visibleBuffs = filterIcons(auras, unit, auras.buffFilter or auras.filter or 'HELPFUL', math.min(numBuffs, max), nil, 0, true)
+		local visibleBuffs = filterIcons(
+			auras,
+			unit,
+			auras.buffFilter or auras.filter or 'HELPFUL',
+			math.min(numBuffs, max),
+			nil,
+			0,
+			true
+		)
```
p3lim, line 390: Same problem here. I'm going to ignore further repetitions of this type of issue.

elements/castbar.lua
```diff
@@ -142,32 +144,49 @@ local function CastStart(self, event, unit)
...
 		safeZone:ClearAllPoints()
 		safeZone:SetPoint(isHoriz and 'TOP' or 'LEFT')
 		safeZone:SetPoint(isHoriz and 'BOTTOM' or 'RIGHT')
 
-		if(element.casting) then
-			safeZone:SetPoint(element:GetReverseFill() and (isHoriz and 'LEFT' or 'BOTTOM') or (isHoriz and 'RIGHT' or 'TOP'))
+		if element.casting then
+			safeZone:SetPoint(
+				element:GetReverseFill() and (isHoriz and 'LEFT' or 'BOTTOM') or (isHoriz and 'RIGHT' or 'TOP')
+			)
```
p3lim, line 174: What is the point of this.

p3lim (commented):
While there are a lot of things to like about this, my main gripes is the unneccessary splitting of args into newlines (I understand that it gives _some_ readability where it lets you easily count the arg index based on the line number), and splitting conditionals of a certain complexity into multiple lines (which in a few cases actually made them less readable).

I have not looked into the configuration possible for this project yet however, these things could possibly be tweaked for all I know.

elements/auras.lua
Rainrider: This one is because the condition exceeded the max line length of 120 cols.

I do find those quite readable btw as it sorts the sub-conditions by precedence which is quite nice. It makes `A or B or C` quire obvious in this case.

elements/auras.lua
Rainrider: This is also because of exceeding 120 cols. I actually like it as well.

elements/castbar.lua
Rainrider: Adhering to line length limit.

elements/auras.lua
Rainrider: There is a plus here - counting arguments is way easier because once can subtract the line numbers :)
But I see what you mean.

PR #613, opened 2022-09-18
Dragonflight

p3lim:
This should contain most, if not all of the fixes and features for Dragonflight.
We'll also merge in any other branches from the [11 milestone](https://github.com/oUF-wow/oUF/milestone/12) if time allows it.

### DIFF
--- a/elements/auras.lua
+++ b/elements/auras.lua
@@ -231,7 +231,7 @@ local function updateIcon(element, unit, index, offset, filter, isDebuff, visibl
 			end
 
 			if(button.icon) then button.icon:SetTexture(texture) end
-			if(button.count) then button.count:SetText(count > 1 and count) end
+			if(button.count) then button.count:SetText(count > 1 and count or '') end
 
 			local width = element.width or element.size or 16
 			local height = element.height or element.size or 16
--- a/elements/auras.lua
+++ b/elements/auras.lua
@@ -406,10 +406,12 @@ local function UpdateAuras(self, event, unit, updateInfo)
 					end
 				end
 			end
+		end
 
-			-- instead of removing auras one by one, just wipe the tables entirely
-			-- and repopulate them, multiple table.remove calls are insanely slow
+		if(buffsChanged or debuffsChanged) then
 			if(buffsChanged) then
+				-- instead of removing auras one by one, just wipe the tables entirely
+				-- and repopulate them, multiple table.remove calls are insanely slow
 				auras.sortedBuffs = table.wipe(auras.sortedBuffs or {})
 
 				for _, data in next, auras.activeBuffs do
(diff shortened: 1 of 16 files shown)


elements/auras.lua
```diff
@@ -422,17 +508,96 @@ local function UpdateAuras(self, event, unit)
...
+				end
+
+				table.sort(buffs.sorted, buffs.SortBuffs or buffs.SortAuras or SortAuras)
+			end
+		end
 
-		local fromRange, toRange
-		if(buffs.PreSetPosition) then
-			fromRange, toRange = buffs:PreSetPosition(numBuffs)
+		if(not buffsChanged) then return end
```
brendanblackwood, line 589: I believe this is causing `UpdateAuras` to exit early if there have been no buff changes even if there have been debuff changes, resulting in debuffs not being added or removed.

ls-: Ehm, it's in `.Buffs`, so obv it only handles buffs...

brendanblackwood: I could totally be misunderstanding how it's supposed to be used, but it appears that if a layout has set both `.Buffs` and `.Debuffs`, it relies on `UpdateAuras` to update both of them, not just the buffs. In this case, we are returning from `UpdateAuras` if there are no changes to buffs even if there are changes to debuffs.

ls-: Ah, true >_> Good catch. More work 😒🔫

p3lim:
TODO (according to @ls-):
- [ ] castbars needs some tweaking
- [ ] docs needs updating
- [ ] merge #592
- [ ] more testing (we should have an example layout that covers everything tbh)

ls-:
Code-wise I think we're ready. Just to clarify on a couple of things.

- I chose against implementing cooldowns for the `Essence` classpower. The way Blizz implemented them is honestly wack. so overcomplicating the element to support that wasn't worth it. If you want them, use `PostUpdate` and `GetPowerRegenForPowerType` to show the "cooldown".

- Castbar empowerment stages. We only add pips to the element, if you need fancy colouring like in the default UI, use those pips as anchors, that's what the default UI does as well. Both horizontal and vertical bars are supported, the reverse fill is supported as well.

- Despite how complex `auras` changes might look, the way the element works now is pretty straightforward. Also, its performance isn't an issue, on average (full and partial updates combined) the new implementation is 30% faster than what we had previously, and even that wasn't slow either. For instance, here's the avg performance per call after 1k calls with like 24-32 active auras:
image

So what's left is to update the documentation where it's outdated or missing.

p3lim:
Posted upgrade guide: https://www.wowinterface.com/forums/showthread.php?t=59273

ls- (dismissed):
👍🏼

ls- (approved):
👍🏼

PR #616, opened 2022-10-25
auras: Add isPlayerAura to data

ls-:
Who could of thunk that `isFromPlayerOrPlayerPet` actually meant that the aura is from ANY player or pet, not just the player. Why isn't it called `isFromAnyPlayerOrPlayerPet`?


p3lim (approved):
LGTM, assuming `UnitIsOwnerOrControllerOfUnit` works for vehicles

ls-:
It does, that's the first thing I tested in the Argent Tournament.

PR #617, opened 2022-10-26
blizzard: Properly disable party frames

ls-:
Blizz reworked how their party frames work, they now use pools. Boss frames still work the old way, btw, no need to change anything there.


p3lim (approved):
LGTM, assuming you've tested this.

PR #621, opened 2022-11-12
auras: element fixes and additions

myzb:
PR as draft, not the lua expert. Feel free to use as inspiration or take what you feel would be beneficial. Tried to keep my additions, fixes and wishes as separate commits for easier cherry-picking.

ead3964b2fea1e74bd9ec462a3c514d00058e032 fixes what we were talking about, with having to track more auras than we are actually showing. Else when one of the showing auras disappear we won't have a aura to fill the gap. 

To test:
1. create a aura element with a small aura.num, say 3-4.
2. Buff yourself to like 4+ buffs
3. click of one the buffs that are currently being shown in the aura element
4. notice that even though there are still 4+ buffs active in the blizzard default player buff bar
5. The empty spot in your aura element is not filled by another aura

### DIFF
--- a/elements/auras.lua
+++ b/elements/auras.lua
@@ -275,13 +275,14 @@ local function UpdateAuras(self, event, unit, updateInfo)
 
 	local auras = self.Auras
 	if(auras) then
-		--[[ Callback: Auras:PreUpdate(unit)
+		--[[ Callback: Auras:PreUpdate(unit, updateInfo)
 		Called before the element has been updated.
 
-		* self - the widget holding the aura buttons
-		* unit - the unit for which the update has been triggered (string)
+		* self       - the widget holding the aura buttons
+		* unit       - the unit for which the update has been triggered (string)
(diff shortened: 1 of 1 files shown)


elements/auras.lua
```diff
@@ -441,14 +412,15 @@ local function UpdateAuras(self, event, unit, updateInfo)
...
+				for i = 1, count do
 					updateAura(auras, unit, auras.sortedBuffs[i], i)
 				end
 			end
 
-			local offset = #auras.sortedBuffs
+			local offset = count
 
-			if(auras.gap and #auras.sortedDebuffs > 0) then
+			if(auras.gap and count > 0) then
```
myzb, line 433: The gap is now always shown though, if buffs are also shown. Since it's an invisible button I guess it does not hurt, otherwise I'd have to move the Show/Hide() call to after updating the debuffs.

ls-:
I can see what you're trying to achieve, but that's def not the best way of doing it. Truth be told, I already had a roughly coded version of that locally. I'll try to PR it next week, so keep your eyes pealed for further testing. Rn I'm taking a break because last 6 weeks were pretty rough on me, so I'm trying to avoid coding as much as possible.

PR #622, opened 2022-11-12
Fix ComboFramePlayer error on druid login

trincasidra:
Fix the following error:
```
Message: Interface/FrameXML/ComboFramePlayer.lua:17: Usage: local power = UnitPower(unitToken [, powerType, unmodified])
Time: Sat Nov 12 14:09:39 2022
Count: 1
Stack: Interface/FrameXML/ComboFramePlayer.lua:17: Usage: local power = UnitPower(unitToken [, powerType, unmodified])
[string "=[C]"]: ?
[string "=[C]"]: in function `UnitPower'
[string "@Interface/FrameXML/ComboFramePlayer.lua"]:17: in function `UpdatePower'
[string "@Interface/FrameXML/ClassPowerBar.lua"]:58: in function `OnEvent'
[string "@Interface/FrameXML/ClassResourceBarTemplate.lua"]:58: in function <Interface/FrameXML/ClassResourceBarTemplate.lua:38>

Locals: 
```

Which seems to be related to druids having a new/different blizzard combo frame.

I also tried tinkered with `handleFrame(ComboPointPlayerFrame)`, but that doesn't seem to add any benefit. Rogues work okay even without that on my implementation at least.

### DIFF
--- a/blizzard.lua
+++ b/blizzard.lua
@@ -76,6 +76,7 @@ function oUF:DisableBlizzard(unit)
 
 	if(unit == 'player') then
 		handleFrame(PlayerFrame)
+		handleFrame(ComboPointDruidPlayerFrame)
 
 		-- For the damn vehicle support:
 		PlayerFrame:RegisterEvent('PLAYER_ENTERING_WORLD')


trincasidra:
Looks like I was too quick - the error still pops up after changing between druid specs.

ls-:
I think I know what's broken and where because it's me who broke it, probably... 😒🔫

ls-:
Hmm, I couldn't reproduce it on my end, and I tried all kinds of player/target frame combinations. What I thought I broke some time ago isn't actually broken. Are you sure you aren't doing something weird on your end? Can I see how you use it?

trincasidra:
Thanks @ls-! You can see how I'm using oUF in my [repo for trincaui](https://github.com/trincasidra/trincaui).

When debugging I found out this error only popped when calling oUF:Spawn for the player frame (core/functions.lua:36 on my repo). When replacing that with a simple CreateFrame I didn't get the error, but of course many things didn't work either, so I went from there to check into the oUF code.

trincasidra:
Great! Error is gone :)

PR #624, opened 2022-11-15
auras: Update element

ls-:
This one addresses #620 and #621 and some other concerns brought up in [our subforum on wowi](https://www.wowinterface.com/forums/showthread.php?t=59334).

I updated the update process so that it properly fills in the blank spots when possible. There's new `.all`, `.allBuffs`, and `.allDebuffs` to keep track of all auras that aren't filtered out by `.filter`, `.buffFilter`, or `.debuffFilter`. `.active` tables no longer carry any aura data, that's what `.all` is for, `.active`s are just lists of active auras. `.sorted` tables are unchanged.

I also added the new `.redrawIfVisibleChanged` option. In some cases, for instance, nameplates users want to redraw the buttons whenever the number of visible buttons changes, this option will allow them to do exactly that.

### DIFF
--- a/elements/auras.lua
+++ b/elements/auras.lua
@@ -480,7 +480,7 @@ local function UpdateAuras(self, event, unit, updateInfo)
 
 			local offset = #auras.sortedBuffs
 
-			if(auras.gap and #auras.sortedDebuffs > 0) then
+			if(auras.gap and offset > 0 and #auras.sortedDebuffs > 0) then
 				offset = offset + 1
 
 				local button = auras[offset]
--- a/elements/auras.lua
+++ b/elements/auras.lua
@@ -15,23 +15,24 @@ At least one of the above widgets must be present for the element to work.
 
 ## Options
 
-.disableMouse       - Disables mouse events (boolean)
-.disableCooldown    - Disables the cooldown spiral (boolean)
-.size               - Aura button size. Defaults to 16 (number)
-.width              - Aura button width. Takes priority over `size` (number)
-.height             - Aura button height. Takes priority over `size` (number)
-.onlyShowPlayer     - Shows only auras created by player/vehicle (boolean)
-.showStealableBuffs - Displays the stealable texture on buffs that can be stolen (boolean)
-.spacing            - Spacing between each button. Defaults to 0 (number)
-.['spacing-x']      - Horizontal spacing between each button. Takes priority over `spacing` (number)
-.['spacing-y']      - Vertical spacing between each button. Takes priority over `spacing` (number)
-.['growth-x']       - Horizontal growth direction. Defaults to 'RIGHT' (string)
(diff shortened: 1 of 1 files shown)


ls-:
K, just pushed prob the biggest change to the new auras.

From now on aura info processing and sorting are completely uncapped. oUF will process all auras on the unit, and trimming to the final number will be happening during the redrawing process. The problem with the current (in master) setup is that we can actually bail out before even reaching important stuff if it's somewhere at the end of the list.

I still chose to keep both `.all*`. and `.active*` tables because I know that many popular layouts have a feature to consolidate auras with the same ID into one button. The most common example is bolstering, it has multiple aura instances for each stack. And to do something like that you'll need the data from `.all*` while keeping only one copy of bolstering in `.active*`. And to check what's in `.all*` and adjust `.active` even further before things proceed to sorting and redrawing there's `PostUpdateInfo`. Due to the nature of partial updates, it's harder to use `FilterAura` for stuff that may rely on other auras being or not being present. Although you still have access to `.all*` and `.active*` from `FA`, so there's a way, it's just way less convenient and prob less efficient as well.

Atm, I have no plans to add aura consolidation to oUF.

myzb:
I like the latest iteration.  I'll test for a few days, but so far so good. Thank you!

I have one last wish. Could it be possible to add the info on whether the update is a full update to `PreUpdate` ? E.g.:

```
local fullupdate = not updateInfo or updateInfo.isFullUpdate
if(auras.PreUpdate) then auras:PreUpdate(unit, fullupdate)
```

Would be helpful so oUF wiping some tables doesn't catch us with our pants down. Also handy to let us know that we probably should also be doing some cleaning up of our own.

ls-:
Hmm, sure...

elements/auras.lua
```diff
@@ -273,15 +274,18 @@ end
 local function UpdateAuras(self, event, unit, updateInfo)
 	if(self.unit ~= unit) then return end
 
+	local isFullUpdate = not updateInfo or updateInfo.isFullUpdate
+
 	local auras = self.Auras
 	if(auras) then
 		--[[ Callback: Auras:PreUpdate(unit)
```
myzb, line 281: small nitpick `isFullUpdate` missing in docstring

p3lim: That field is provided by the `UNIT_AURA` event, not this element.

p3lim: Oh you mean the field for PreUpdate? That is already in the docstring.

myzb: the description: PreUpdate(unit, isFullUpdate)

ls-:
Added `element:PostProcessAuraData(unit, data)` for when you need to tweak the data, but be careful with it.

p3lim (approved):

PR #625, opened 2022-11-15
Investigate switching to UPU from UPF for class powers

ls-:
UPF is a very spammy event, and using it for something like class powers feels like an overkill. They simply don't continuously regen like , let's say, rogue's Energy. The only reason why we use UPF is because it's been this way for over a decade at this point 😒

I'm not sure if there's any drawbacks to using UPU hence the name of the PR. I'll switch my own UI to UPU to gather as much feedback as possible from its users, but help from others is always welcome 😊

### DIFF
--- a/elements/classpower.lua
+++ b/elements/classpower.lua
@@ -255,7 +255,7 @@ end
 do
 	function ClassPowerEnable(self)
 		self:RegisterEvent('UNIT_MAXPOWER', Path)
-		self:RegisterEvent('UNIT_POWER_FREQUENT', Path)
+		self:RegisterEvent('UNIT_POWER_UPDATE', Path)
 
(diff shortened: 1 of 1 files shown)


Wutname1:
This should be good to go right?

Haven't heard of any issues with it and it's been running smoothly in SpartanUI since the Nov 2022 commit. Also, Elv added it last March and it made it into their v13.28 release in mid-March 2023.

arnvid:
Been using this in RealUI also for a while now

ls-:
Yeah, I've been using it in my UI for over a year now as well, that's why I keep bumping this branch 😅
I guess it's time to ask the guys to approve it.

Rainrider (approved):

PR #635, opened 2023-03-07
tags: Update element

ls-:
The element desperately needed some love.

The main goal of this PR was to increase the performance of the element. Tags were originally written with standard unit frames in mind, and they're fairly static. But with the introduction of the nameplates some performance issues reared their ugly heads. Unfortunately, it took me a while to get to this because everything was... still working 😅

There's generally two areas of improvement:
- **Staggered updates.** Each tag string usually has multiple events registered for its updates, and quite often they fire in rapid succession while the data that's being displayed stays the same. So instead of updating a tag string each time any of its events fires, we just mark it for an update when that happens, but we'll update it only once every 0.25s or 4 times a second. You'll be able to adjust the update frequency, but the minimum is 0.1s. But even 4 times a second is more than enough, IME. This allows us to drop around 3-6k unnecessary updates during a 5 minute world boss fight.
- **Faster tag un/registration.** The one is the heavy hitter. The old tag unregistration process in particular was slow because it's written with fairly static standard unit frames in mind. We just didn't need to unregister and register tags again and again and again whenever frames were shown and hidden because, let's say, a target frame was always watching a `"target"` unit. But a nameplate gets a different unit assigned to it whenever its shown on the screen, one moment it's `"nameplate1"`, but the next time you see it it's `"nameplate26"`. The new tag re-registration process is fast, that's primarily where all the performance gains are coming from.

The overall performance of the element was increased by **OVER 10 TIMES** 🥳 Again, not 10%, but 10 times! 🤯 It'll be particularly noticeable in the areas where you get a lot of nameplates. How bad was that? Let's say, if you had 15-25 nameplates on the screen you could drop **up to** 20% of your FPS while turning/spinning your camera really fast. Yikes! 😨

I also cleaned the code up a bit, added data caching here and there, and gave various variables more self-explanatory and consistent names ☺

To adjust the update frequency, use `oUF.Tags:SetEventUpdateTimer(timer)`.

### DIFF
--- a/elements/tags.lua
+++ b/elements/tags.lua
@@ -769,9 +769,9 @@ local function getTagFunc(tagstr)
 					tagFunc = function()
 						return '|cffffffff' .. bracket .. '|r'
 					end
-				end
 
-				invalidBrackets[bracket] = tagFunc
+					invalidBrackets[bracket] = tagFunc
+				end
 			end
 
 			table.insert(funcs, tagFunc)
--- a/elements/tags.lua
+++ b/elements/tags.lua
@@ -108,7 +108,7 @@ local Private = oUF.Private
 
 local nierror = Private.nierror
 local unitExists = Private.unitExists
-local xpcall = Private.xpcall
+local validateEvent = Private.validateEvent
 
 local _PATTERN = '%[..-%]+'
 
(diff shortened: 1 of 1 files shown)


Rainrider (approved):
LGTM

Rainrider (approved):

PR #653, opened 2023-07-14
blizzard: Hide the new compact arena frames

p3lim:
Ref: https://github.com/Gethe/wow-ui-source/blob/7721fd02dc9425ea89893807f43eaaaa4b7fdd0a/Interface/FrameXML/CompactArenaFrame.lua#L78-L88


ls- (approved):
LGTM👌🏼

PR #654, opened 2023-07-18
castbar: Improve empowered cast handling

ls-:
Empowered casts needed some love.

First of all, I added missing `PostUpdatePips` callback, users basically had no other option, but to override `CreatePips` and then jump through the hoops in `PostCastStart`.

Secondly, I added `PostUpdateStage` to notify the layout whenever the current stage changes.

### DIFF
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -173,6 +173,16 @@ local function UpdatePips(element, numStages)
 			end
 		end
 	end
+
+	--[[ Callback: Castbar:PostUpdatePips(unit)
+	Called after the element has updated stage separators (pips) in an empowered cast.
+
+	* self - the Castbar widget
+	* numStages - the number of stages in the current cast (number)
+	--]]
+	if(element.PostUpdatePips) then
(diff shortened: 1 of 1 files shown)


elements/castbar.lua
```diff
@@ -117,6 +121,8 @@ local function UpdatePips(element, numStages)
 	local stageMaxValue = element.max * 1000
 	local isHoriz = element:GetOrientation() == 'HORIZONTAL'
 	local elementSize = isHoriz and element:GetWidth() or element:GetHeight()
+	element.numStages = numStages
+	element.curStage = 0 -- NOTE: Updates only if the PostUpdateStage callback is present
```
p3lim, line 125: missing docs for these two

p3lim (changes requested):

elements/castbar.lua
ls-: You mean like adding them to the attributes list?

ls-: 😖

p3lim (approved):

PR #655, opened 2023-07-30
adding project id variables

pedrozc90:
Hello.

Do you mind to keep these variables to track the game expansion?
This change will help a lot, enabling me to use oUF repository as a git submodule.


ls-:
Why? Officially, we only support retail since none of us play classic. ElvUI guys added support for classic versions on their own, and they have what you suggest in their implementation.

PR #656, opened 2023-09-06
core: Support contextual pings

p3lim:
They can be disabled in other ways than setting `frame.IsPingable = false`, but it's the simplest way (other methods involve attributes).


Rainrider (approved):
lgtm

PR #659, opened 2023-09-20
core: Fix contextual ping on suffix units

p3lim:

PR #661, opened 2023-10-31
core: Fix contextual ping on header children

p3lim:
When header children spawn during combat (e.g. a raid member joins) it would throw an error because we can't set attributes then.

Moving the attribute setter into a non-header block plus `initialConfigFunction` solves this, while we gotta keep the Mixin call where it was as the Mixin function nor the arg we pass through it exists in a secure environment.

Ref https://github.com/oUF-wow/oUF/pull/659#discussion_r1375492483


Rainrider (approved):
LGTM

Meivyn:
Confirm that this fixed the issue.

PR #685, opened 2024-07-31
core: Add support for passing header attributes as a table

p3lim:


ouf.lua
```diff
@@ -640,10 +640,19 @@ do
...
 
 		header:SetAttribute('template', 'SecureUnitButtonTemplate, SecureHandlerStateTemplate, SecureHandlerEnterLeaveTemplate, PingableUnitFrameTemplate')
-		for i = 1, select('#', ...), 2 do
-			local att, val = select(i, ...)
-			if(not att) then break end
-			header:SetAttribute(att, val)
+
+		if(...) then
+			if(type((...)) == 'table') then
+				for att, val in next, (...) do
```
ls-, line 646: You overdid it with parens here, `...` is totally fine..

ouf.lua
```diff
@@ -640,10 +640,19 @@ do
 		local header = CreateFrame('Frame', name, PetBattleFrameHider, template)
 
 		header:SetAttribute('template', 'SecureUnitButtonTemplate, SecureHandlerStateTemplate, SecureHandlerEnterLeaveTemplate, PingableUnitFrameTemplate')
-		for i = 1, select('#', ...), 2 do
-			local att, val = select(i, ...)
-			if(not att) then break end
-			header:SetAttribute(att, val)
+
+		if(...) then
+			if(type((...)) == 'table') then
```
ls-, line 645: And here... Just `type(...)` should do the trick.

ls- (approved):
🫡

PR #694, opened 2024-11-20
power: Add support for atlas textures

p3lim:


ls- (commented):
LGTM 👌🏼

p3lim:
This needs to revert to the previously used texture when switching back to _not_ using an atlas, otherwise the atlas texture would be used for colors (e.g. switching between moonkin form and cat form as a balance druid).

p3lim:
Also think the option name should be more verbose, maybe `colorPowerAtlas`?

ls-:
No. Why? It's on the `.Power` element. `frame.Power.usePowerAtlas` is overly verbose.

elements/power.lua
```diff
@@ -160,31 +165,40 @@ local function UpdateColor(self, event, unit)
...
+			r, g, b = color[1], color[2], color[3]
+		end
 
-	if(b) then
-		element:SetStatusBarColor(r, g, b)
+		if(b) then
+			if(element.__texture) then
+				element:SetStatusBarTexture(element.__texture)
+			end
+			element:SetStatusBarColor(r, g, b)
```
ls-, line 180: Need an empty line after the if block.

p3lim:
> No. Why? It's on the .Power element. frame.Power.usePowerAtlas would be overly verbose.

Since this option is essentially the equivalent to coloring the status bar it makes sense that the option is named as such.

ls-:
> Since this option is essentially the equivalent to coloring the status bar it makes sense that the option is named as such.

Nah.

ls- (dismissed):
LGTM 👌🏼

elements/power.lua
```diff
@@ -160,31 +167,41 @@ local function UpdateColor(self, event, unit)
...
-	* unit - the unit for which the update has been triggered (string)
-	* r    - the red component of the used color (number)[0-1]
-	* g    - the green component of the used color (number)[0-1]
-	* b    - the blue component of the used color (number)[0-1]
+	* self  - the Power element
+	* unit  - the unit for which the update has been triggered (string)
+	* r     - the red component of the used color (number?)[0-1]
+	* g     - the green component of the used color (number?)[0-1]
+	* b     - the blue component of the used color (number?)[0-1]
+	* atlas - the atlas used instead of color (string)
```
ls-, line 201: I reckon it should be `string?` just like `number?` since it's not guaranteed to be there.

elements/power.lua
```diff
@@ -146,6 +149,10 @@ local function UpdateColor(self, event, unit)
 		else
 			color = self.colors.power[ALTERNATE_POWER_INDEX]
 		end
+
+		if(element.colorPowerAtlas and color and color.atlas) then
```
ls-, line 153: The `color.atlas` check is unnecessary, it being `nil` isn't a problem.

elements/power.lua
```diff
@@ -398,6 +415,10 @@ local function Enable(self)
 			element:SetStatusBarTexture([[Interface\TargetingFrame\UI-StatusBar]])
 		end
 
+		if(element.colorPowerAtlas) then
+			element.__texture = element:GetStatusBarTexture():GetTexture()
```
ls-, line 419: `element.__texture = element.__texture or element:GetStatusBarTexture():GetTexture()` might be a better idea in case it's not the initial `Enable` call because the bar might be using an atlas from before. So trying to recover it from `element.__texture` is safer, and users might set it themselves to avoid any inconsistencies as well.

ls- (approved):
LGTM 👌🏼

PR #695, opened 2024-11-20
power: Store and forward the used minimum value correctly

p3lim:


ls- (commented):
`min = min or 0` seems more concise than this if block?

p3lim:
> `min = min or 0` seems more concise than this if block?

Sure.

ls- (approved):
LGTM 👌🏼

PR #699, opened 2024-11-21
core: Nameplate sub-widgets should ignore alpha

p3lim:
It's a common practice to hide nameplates by adjusting their alpha, but this has adverse effects on blizzard-managed widgets, affecting gameplay in an (usually) accidental way.


ls- (approved):
LGTM 👌🏼

PR #700, opened 2024-11-21
totems: Follow priorities used by Blizzard

p3lim:
Ref: https://github.com/Gethe/wow-ui-source/blob/fa3531505e3e2406f9e3d5a0fbd9644bd5c85227/Interface/AddOns/Blizzard_UnitFrame/TotemFrame.lua#L15-L18


ls- (approved):
LGTM 👌🏼

PR #702, opened 2024-11-25
totems: Use 4 elements in the example

p3lim:
In https://github.com/oUF-wow/oUF/commit/c08b0d7a954b20ea55883ebafdefe7ecc6c1ea8d the constant was replaced with `5`, even though it has always represented `4`.

Mentioned in #701.


ls- (approved):
LGTM 👌🏼

PR #707, opened 2025-02-16
Update power.lua

dwWei:
This can fix the issue where the energy bar sometimes doesn't change when team members go offline.


p3lim (approved):

p3lim:
Feel free to do the same with other elements, including the name tag, as they would benefit from this too.

PR #725, opened 2025-10-24
Midnight WIP

p3lim:
Most of these tests are being performed while in combat inside a dungeon, to ensure compatibility with secrets.

## Legend:

- :green_circle: = everything works just fine as-is or with changes in this PR
- :yellow_circle: = mostly works or can be made to work with changes, or needs further testing
- :red_circle: = something is broken and we can't fix it
  - if it has this mark then we expect this to work but need Blizzard to do something
- :skull: = completely dead and is intentionally not going to work
- if something doesn't have an emoji then it just means I haven't tested it yet, don't nag, be patient

## Status of checked things (last check = build 65028 (PTR RC 2)):

- :green_circle: core
- :green_circle: standard unit logic (i.e. player, target, etc)
- :green_circle: groups
  - they spawn atleast, need to check what happens when players join/leave in combat
- :green_circle: nameplates
  - they work fine, but we have to deploy a hack to make them clickable
  - this hack will not be necessary once they add a hitrect reparent API (planned)
- arena prep

## Elements:

- additionalpower
- alternativepower
- :green_circle: assistantindicator
  - only tested on player+target
- :green_circle: auras
    - :green_circle: dispel color
    - :green_circle: "stealable"
    - :green_circle: cooldown spiral
    - :green_circle: aura count
    - :green_circle: tooltips
    - :yellow_circle: sorting by time
      - there are new APIs for sorting, but it'll require a non-insignificant rework of how we deal with UNIT_AURA payload, so it'll be TBD for a bit
- :yellow_circle: castbar
  - :green_circle: cast; statusbar, text, time, icon
  - :green_circle: channel
  - :green_circle: tracking casts across events
  - :green_circle: empowered abilities for non-player
    - :red_circle: unable to determine which stage is currently active, which makes fancy stuff less fancy
  - :red_circle: cast/channel delay for non-player
    - planned new secrets formatter specifically for this in a later build
  - :green_circle: change color based on interruptible state
  - :green_circle: tradeskill hiding
- :green_circle: classpower
- :green_circle: combatindicator
  - only tested on player
- :green_circle: grouproleindicator
- :green_circle: health
- :green_circle: healthprediction
- :green_circle: leaderindicator
  - only tested on player+target
- phaseindicator
- portrait
- :green_circle: power
- :green_circle: powerprediction
  - only tested against power, not additionalpower
- pvpclassificationindicator
- :green_circle: pvpindicator
  - only tested on player
- questindicator
- raidroleindicator
- :green_circle: raidtargetindicator
  - not tested on group/nameplates yet
- :green_circle: range
- :green_circle: readycheckindicator
- :green_circle: restingindicator
  - only tested on player
- resurrectindicator
- :green_circle: runes
- :green_circle: stagger
- summonindicator
- :yellow_circle: tags
  - when a tag contains secrets we're limited with what we can check (i.e. if a tag is empty or not), and I assume any boolean APIs would return secrets too so that won't fix anything
  - because of this it might be better (for layouts) to deal with updating fontstrings through PostUpdate of relevant elements (like before the tags were added) or with custom event handlers (which oUF makes pretty straight forward to add)
  - STATUS: need to decide what to do with this element
- threatindicator
- totems

## New stuff:

- Demon Hunter soul fragments
  - uses UNIT_AURA for some dumb reason, so can't put it in classpower

### DIFF
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -35,6 +35,7 @@ A default texture will be applied to the StatusBar and Texture widgets if they d
 .empowering       - Indicates whether the current spell is an empowering cast (boolean)
 .notInterruptible - Indicates whether the current spell is interruptible (boolean)
 .spellID          - The spell identifier of the currently cast/channeled/empowering spell (number)
+.spellName        - The name of the spell currently being cast/channeled/empowered (string)
 
 ## Examples
 
@@ -96,6 +97,7 @@ local function resetAttributes(self)
 	self.empowering = nil
 	self.notInterruptible = nil
 	self.spellID = nil
+	self.spellName = nil
 
 	for _, pip in next, self.Pips do
 		pip:Hide()
@@ -222,6 +224,7 @@ local function CastStart(self, event, unit)
 	element.holdTime = 0
 	element.castID = castID
 	element.spellID = spellID
+	element.spellName = text
 
 	if(unit == 'player') then
 		-- we can only read these variables for players


ls-:
giphy

remyeik:
Vel blåst haste! ❤️

GuillaumeEllby:
Really, sad I have been using oUF since OG WoTLK, I really hope they fix this so I still use oUF after all these years.

myzb:
Just wanted to drop this here as a big thank you! 
This pic is special, Nzoth was my first cutting edge and the first boss I rolled a self-made UI based on oUF. :heart: 

WoWScrnShot_052820_220322

ls-:
Because this PR and our comments here are making the rounds online rn, I'll repost my ["rant" from reddit](https://www.reddit.com/r/WowUI/comments/1ofqb0p/comment/nlb1ofa/) here to provide more context.

image

p3lim:
Leaving player secondary power tags as-is for now, as Blizzard has said they'd be made non-secret later in the beta.

p3lim:
As for [85aba05](https://github.com/oUF-wow/oUF/pull/725/commits/85aba05b4a9f96264733fbb7e933b4760d85d703), `:SetShown(secretBoolean)` should also be possible in a later beta build.

That'll solve the `stealable` logic.

ls-:
> Demon Hunter soul fragments
>
>    uses UNIT_AURA for some dumb reason, so can't put it in classpower

Aug evoker bar also works similarly. It's called ebon might, and it's basically a fancy way of showing the duration of the buff. And for DHs it's just a way of showing stacks on the buff.

Imho, these things should be in the cooldown managers, not these fake classpower bars.

Witnesscm:
Hi, I’m not entirely sure if this is the right place to report this.

I just tested on build 64774, the unit frames created with oUF (only health bar and no other elements) fail to show and hide properly during combat. And the client throws  blocked error. At first I suspected something was wrong with Blizzard’s RegisterUnitWatch, but when I tested with small standalone snippet that creates a basic target frame, that frame behaved correctly and did not trigger any blocked actions.

p3lim:
@Witnesscm the restricted environment is just broken in this build.

colors.lua
```diff
@@ -129,9 +129,14 @@ if(not customClassColors()) then
...
-	colors.debuff[debuffType] = oUF:CreateColor(color.r, color.g, color.b)
-end
+-- copy of DEBUFF_DISPLAY_INFO from AuraUtil
+colors.debuff[oUF.Enum.DebuffType.None] = _G.DEBUFF_TYPE_NONE_COLOR
+colors.debuff[oUF.Enum.DebuffType.Magic] = _G.DEBUFF_TYPE_MAGIC_COLOR
+colors.debuff[oUF.Enum.DebuffType.Curse] = _G.DEBUFF_TYPE_CURSE_COLOR
+colors.debuff[oUF.Enum.DebuffType.Disease] = _G.DEBUFF_TYPE_DISEASE_COLOR
+colors.debuff[oUF.Enum.DebuffType.Poison] = _G.DEBUFF_TYPE_POISON_COLOR
+colors.debuff[oUF.Enum.DebuffType.Bleed] = _G.DEBUFF_TYPE_BLEED_COLOR
+colors.debuff[oUF.Enum.DebuffType.Enrage] = CreateColor(243/255, 95/255, 245/255) -- TODO: make something up
```
ls-: It's a semantic/naming issue, but it should be `dispel` and not `debuff`. Enrage is not a debuff, it's a buff.

ls- (changes requested):
Went through most of the changes, but I noticed a couple of naming issues here and there. In a few areas what's called `debuff` should be `dispel`.

elements/castbar.lua
```diff
@@ -314,53 +295,51 @@ local function CastStart(self, event, unit)
...
+	if(not element:IsShown() or element.castID ~= castID) then
 		return
 	end
 
-	local name, startTime, endTime, _
+	local direction, duration, name, startTime, _ = Enum.StatusBarTimerDirection.ElapsedTime
 	if(event == 'UNIT_SPELLCAST_DELAYED') then
-		name, _, _, startTime, endTime = UnitCastingInfo(unit)
+		name, _, _, startTime = UnitCastingInfo(unit)
+		duration = UnitEmpoweredChannelDuration(unit)
```
ls-: I reckon it should be `UnitCastingDuration` here.

elements/castbar.lua
```diff
@@ -455,77 +484,46 @@ end
...
-						end
-					else
-						break
-					end
-				end
-			end
-		end
-
-		self:SetValue(self.duration)
+		-- --[[ Callback: Castbar:PostUpdateStage(stage)
```
ls-: I'd add why it's commented out atm.

elements/castbar.lua
p3lim: No this is intentional, this duration API specific to empowered casts includes the HoldAtMax endtime, something `UnitCastingDuration` doesn't do.

Witnesscm:
Thanks for the huge amount of work.

While testing, I noticed a small issue in tags:  
UnitHealthPercent and UnitPowerPercent now need to be called like this:
- UnitHealthPercent(u, true, CurveConstants.ScaleTo100)
- UnitPowerPercent(u, nil, true, CurveConstants.ScaleTo100)

elements/castbar.lua
ls-: Fr? Wtf? So initially we need to use `UnitChannelDuration`, but later on we have to use `UnitEmpoweredChannelDuration`? Even though it's not an empowered cast? This feels like a bug.

ls-: Oof, you spooked me 😅

Rainrider:
I tried to replace `oUF:ColorGradient` by a curve like that:

```lua
local curve = _G.C_CurveUtil.CreateColorCurve()
curve:AddPoint(0.00, _G.CreateColor(0.69, 0.31, 0.31))
curve:AddPoint(0.33, _G.CreateColor(0.71, 0.43, 0.27))
curve:AddPoint(0.66, _G.CreateColor(0.17, 0.17, 0.24))

local function UpdateColor(self, _, unit)
	local r, g, b, color
	local health = self.Health

	if health.colorDisconnected and (not UnitIsConnected(unit) or UnitIsDeadOrGhost(unit)) then
		health:SetValue(health.max or 1)
		color = colors.disconnected
	elseif health.colorTapping and not UnitPlayerControlled(unit) and UnitIsTapDenied(unit) then
		color = colors.tapped
	elseif health.colorSmooth then
		color = UnitHealthPercent(unit, true, curve)
	else
		color = colors.health
	end

	if color then
		r, g, b = color.r, color.g, color.b
	end

	if b then
		health:SetStatusBarColor(r, g, b)
	end
end
```

and it results in

```
62x oUF_Layout/elements/health.lua:41: bad argument #1 to 'SetStatusBarColor' (Usage: self:SetStatusBarColor(color [, a]). Secret values are only allowed during untainted execution for this argument.)
[oUF_Layout/elements/health.lua]:41: in function <oUF_Layout/elements/health.lua:20>
[oUF/elements/health.lua]:182: in function <oUF/elements/health.lua:174>
[oUF/elements/health.lua]:241: in function 'func'
[oUF/ouf.lua]:200: in function <oUF/ouf.lua:183>
```

Am I doing it wrong or is something in the works? Using `oUF:ColorGradient` fails as well, because `health.cur` and `health.max` are secrets too.

ls-:
Then try using your original code, but then replace
```lua
if color then
    r, g, b = color:GetRGB()
end
```
??? It should work in theory 🤔

elements/castbar.lua
```diff
@@ -373,37 +352,78 @@ local function CastUpdate(self, event, unit, castID, spellID)
...
+		if(element.Text) then element.Text:SetText(INTERRUPTED) end
+
+		element.holdTime = element.timeToHold or 0
+
+		-- force filled castbar
+		element:SetMinMaxValues(0, 1)
+		element:SetValue(1)
+	end
+
 	resetAttributes(element)
```
Rainrider: Shouldn't this be done after the post-hooks? Post-hooks cannot get the `spellID` this way.

elements/range.lua
```diff
@@ -45,11 +45,7 @@ local function Update(self, event)
 	local connected = UnitIsConnected(unit)
 	if(connected) then
 		inRange, checkedRange = UnitInRange(unit)
-		if(checkedRange and not inRange) then
-			self:SetAlpha(element.outsideAlpha)
-		else
-			self:SetAlpha(element.insideAlpha)
-		end
+		self:SetAlphaFromBoolean(inRange, element.insideAlpha, element.outsideAlpha)
```
Rainrider: `UnitInRange` returns `false` `false` when not in a group, so the target unit appears always out of range this way.

I would have suggested

```suggestion
		self:SetAlphaFromBoolean(checkedRange and not inRange, element.outsideAlpha, element.insideAlpha)
```

but it results in `attempt to perform boolean test on local 'checkedRange' (a secret value)`.

elements/castbar.lua
ls-: Actually, I'll just move the resetter. Maybe someone might want it for some player-related cleanup in `Post*`s.

Rainrider: > Tbf, spellID is almost always secret anyway, that's why p3lim dropped it from Post*s, it's not even used anywhere in the element either.

```lua
local function PostCastFailedOrInterrupted(castbar)
	castbar:SetStatusBarColor(0.69, 0.31, 0.31)

	local time = castbar.Time
	if time then
		time:SetText(C_Spell.GetSpellName(castbar.spellID))
	end
end
```

works for me. Thanks for the change.

ouf.lua
```diff
@@ -839,32 +901,64 @@ function oUF:SpawnNamePlates(namePrefix, nameplateCallback, nameplateCVars)
...
-			if(nameplateCallback) then
-				nameplateCallback(nameplate.unitFrame, event, unit)
+			if(self.removedCallback) then
+				self.removedCallback(nameplate.unitFrame, event, unit)
 			end
 		end
-	end)
+	end
+
+	--[[ oUF:SpawnNamePlates(prefix, callback, variables)
```
Wetxius: Forgot to remove "callback, variables" from function

ls-:
@arnvid should be fixed, the files were loading in the wrong order.

p3lim:
If you replace any colors you have to use `oUF:CreateColor`.
`r, g, b = color[1], color[2], color[3]` this is also gone.

DaleHuntGB:
Health Prediction might have the incorrect checks for clamping?

```
if(element.damageAbsorbBounds) then
	element.values:SetDamageAbsorbClampMode(element.damageAbsorbBounds)
end

if(element.healAbsorbBounds) then
	element.values:SetHealAbsorbClampMode(element.healAbsorbBounds)
end
```
Should be:
```
if(element.damageAbsorbClampMode) then
	element.values:SetDamageAbsorbClampMode(element.damageAbsorbClampMode)
end

if(element.healAbsorbClampMode) then
	element.values:SetHealAbsorbClampMode(element.healAbsorbClampMode)
end
```
Unless I am misunderstanding?

ls- (dismissed):
LGTM 🙏🏼

DaleHuntGB:
@ls- Ok, I figured that might have been the reason 😔

Wetxius:
> @DaleHuntGB yes, we can't do maths on secret values.

But class color and reaction color not secret and can math :(

ls-:
> We still do something like this
> health.bg:SetVertexColor(r, g, b, 0.2)

And that's exactly why `PostUpdateColor` exists.

p3lim:
You can make them invisible (alpha), but they will still occupy the space and respond to mouse events.

> whitelist/blacklist is still broken

It's not "broken", it's intended, doing spellID filtering will forever be impossible.

The only filtering we can do now is through the AuraFilter option, see this: https://warcraft.wiki.gg/wiki/API_C_UnitAuras.GetAuraDataByIndex

DaleHuntGB:
> You can make them invisible (alpha), but they will still occupy the space and respond to mouse events.

This is what I assumed would be the case. Which is why I thought I would ask. 

> It's not "broken", it's intended, doing spellID filtering will forever be impossible.

Yeah, "broken" was the incorrect wording probably. 

That link is useful, thanks! Appreciate the response!

PR #741, opened 2026-01-18
castbar: Temporary fix bug with event payload

p3lim:
ElvUI team reported something fishy going on with the event payload when casting [Dream Breath](https://www.wowhead.com/spell=355936/dream-breath), where the `UNIT_SPELLCAST_EMPOWER_START` fires twice, the first one containing swapped `spellID` and `castbarID` arg positions (without preserving secret aspect on the values), as present in this screenshot:

image

This PR fixes that for oUF.

Blizzard is aware and will be looking into a fix.

### DIFF
--- a/elements/castbar.lua
+++ b/elements/castbar.lua
@@ -185,6 +185,13 @@ local function ShouldShow(element, unit)
 	return element.__owner.unit == unit
 end
 
+local BROKEN_EMPOWER_SPELLS = {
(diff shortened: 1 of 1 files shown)


p3lim:
Bug confirmed by Blizzard and they've found the cause.
A fix will be deployed _after_ the prepatch release as a hotfix.

PR #765, opened 2026-01-26
core: Disable ping support

p3lim:
This is currently broken in combat, Blizzard is aware but we disable them until it's been fixed.


ls- (approved):
🫡

PR #777, opened 2026-01-29
health: Merge HealthPrediction

### LINKED ISSUE #763: Merge HealthPrediction element into Health
With all of the changes to the HealthPrediction element that happened in 12.0.0 the element now is barely recognizable, and coming in 12.0.1 we're getting more methods added to the "calculator" used to assign all the values to it, including replacements for `UnitHealth` and `UnitHealthMax` which makes the HealthPrediction part of it better (like more clamping options).

The natural thing to do then is to just merge this element with Health, giving it a common API to request health-related data (with temp loss the odd duck out).

We'll see how the calculation API finalizes after the beta and decide then, but it's likely this will happen in v14 of oUF.

p3lim:
Fixes #763

### DIFF
--- a/elements/health.lua
+++ b/elements/health.lua
@@ -502,6 +502,9 @@ local function Enable(self)
 			element.values:SetIncomingHealOverflowPercent(element.incomingHealOverflow)
 		end
 
+		self:RegisterEvent('UNIT_HEALTH', Path)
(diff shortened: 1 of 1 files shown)


ls- (approved):
🫡

PR #778, opened 2026-01-30
aura: sometimes auras comes back as nil. this helps this not to happen.

arnvid:
aura: partial updates can happen before any full update, so make sure caches exist. 
Sometimes this triggered on non existing data.


p3lim:
Should be solved with #783.

PR #783, opened 2026-02-10
auras: Initial update is always full update

ls-:
The first update after the element was (re)enabled should always be a full update regardless of what the payload might suggest.


p3lim (approved):
LGTM

PR #808, opened 2026-02-15
power: Merge in PowerPrediction

p3lim:


elements/power.lua
```diff
@@ -425,6 +551,22 @@ local function Enable(self)
...
+			if(UnitIsUnit(unit, 'player')) then
+				if(element.CostPrediction:IsObjectType('StatusBar') and not element.CostPrediction:GetStatusBarTexture()) then
+					element.CostPrediction:SetStatusBarTexture([[Interface\TargetingFrame\UI-StatusBar]])
+				end
+
+				self:RegisterEvent('UNIT_SPELLCAST_START', PredictionPath)
+				self:RegisterEvent('UNIT_SPELLCAST_STOP', PredictionPath)
+				self:RegisterEvent('UNIT_SPELLCAST_FAILED', PredictionPath)
+				self:RegisterEvent('UNIT_SPELLCAST_SUCCEEDED', PredictionPath)
+				self:RegisterEvent('UNIT_DISPLAYPOWER', PredictionPath)
```
ls-: I'm not entirely sure since I wrote the original ages ago, but I don't think `UNIT_DISPLAYPOWER` is needed at this point. Iirc, it's there because the original powerpred covered both main and additional powers, but this iteration is tied to one specific element.

The same applies to `UDP` in additionalpower, btw.

ls- (approved):
🫡

PR #811, opened 2026-02-15
v14

p3lim:


ls-:
I'll leave a reminder for myself here.

Update `UpdateColor` and `ColorPath` in classpower.lua to match the rest of the elements. Currently they're in the backwards compat mode.

p3lim:
Reminder for myself; go through elements and ensure that internal state attributes attached to widgets are uniformly prefixed with `__`.

p3lim:
Reminder to remove health & power SetColorDisconnected after

p3lim:
Reminder to clean up deprecations in totem element

elements/castbar.lua
```diff
@@ -708,6 +731,10 @@ local function Disable(self)
 
 		element:SetScript('OnUpdate', nil)
 
+		if(element.Time and element.Time.binding) then
```
ls-: Is this `element.Time.binding` check to protect from `element.Time.binding = false` or something? If not, I don't think it's even needed because we create our own if none is defined as long as `.Time` is created.

Actually, the same applies to all of these checks in `CastStart`, `CastUpdate`, etc.

elements/tags.lua
```diff
@@ -851,18 +848,17 @@ Used to register a tag on a unit frame.
...
+	elseif(STATE[self][fs]) then
 		self:Untag(fs)
 	end
 
-	fs.parent = self
+	fs.__owner = self
 	fs.UpdateTag = getTagFunc(ts)
 
-	if(self.__eventless or fs.frequentUpdates) then
+	if(self:IsEventless() or fs.frequentUpdates) then
```
ls-: While reviewing this stuff I remembered an ancient tags' bug I never ended up fixing. Basically, for eventless frames tags end up being updated TWICE because tags' own updater does it, and the eventless frame's updater also does it via `object:UpdateAllElements('OnUpdate')` in `units.lua`. They might be using different update frequencies, but in reality I think for eventless frames `fs.frequentUpdates` should be entirely ignored, and we should only allow it for event driven frames.

elements/castbar.lua
p3lim: I guess the extra check could be removed.

elements/tags.lua
p3lim: Like this? https://github.com/oUF-wow/oUF/pull/811/commits/5b6b6dcc236d512166a178e5144be688138597d2

elements/castbar.lua
p3lim: https://github.com/oUF-wow/oUF/pull/811/changes/32a8632a3b2b9fa9a6af504ac9ec710d9ac92179

elements/tags.lua
ls-: Yup :v

elements/additionalpower.lua
```diff
@@ -185,17 +185,18 @@ end
...
 
 local function shouldUpdatePredictionSize(self)
 	local element = self.AdditionalPower
 
-	local isHoriz = element:GetOrientation() == 'HORIZONTAL'
-	local newSize = element[isHoriz and 'GetWidth' or 'GetHeight'](element)
-	if(isHoriz ~= STATE[element].isHoriz or newSize ~= STATE[element].size) then
-		STATE[element].isHoriz = isHoriz
+	local horizontal = element:GetOrientation() == 'HORIZONTAL'
+	local newSize = element[horizontal and 'GetWidth' or 'GetHeight'](element)
```
ls-: Mmmm, maybe do what you do in castbar.lua?
```lua
local newSize = horizontal and element:GetWidth() or element:GetHeight()
```
?

Looks cleaner, imho.

elements/health.lua
```diff
@@ -304,34 +304,35 @@ end
...
 
 local function shouldUpdatePredictionSize(self)
 	local element = self.Health
 
-	local isHoriz = element:GetOrientation() == 'HORIZONTAL'
-	local newSize = element[isHoriz and 'GetWidth' or 'GetHeight'](element)
-	if(isHoriz ~= STATE[element].isHoriz or newSize ~= STATE[element].size) then
-		STATE[element].isHoriz = isHoriz
+	local horizontal = element:GetOrientation() == 'HORIZONTAL'
+	local newSize = element[horizontal and 'GetWidth' or 'GetHeight'](element)
```
ls-: Same thing here.

elements/power.lua
```diff
@@ -319,17 +319,18 @@ end
...
 
 local function shouldUpdatePredictionSize(self)
 	local element = self.Power
 
-	local isHoriz = element:GetOrientation() == 'HORIZONTAL'
-	local newSize = element[isHoriz and 'GetWidth' or 'GetHeight'](element)
-	if(isHoriz ~= STATE[element].isHoriz or newSize ~= STATE[element].size) then
-		STATE[element].isHoriz = isHoriz
+	local horizontal = element:GetOrientation() == 'HORIZONTAL'
+	local newSize = element[horizontal and 'GetWidth' or 'GetHeight'](element)
```
ls-: And here.

elements/castbar.lua
```diff
@@ -110,8 +110,8 @@ local function CreatePip(element)
 end
 
 local function UpdatePips(element, stages)
-	local isHoriz = element:GetOrientation() == 'HORIZONTAL'
-	local elementSize = isHoriz and element:GetWidth() or element:GetHeight()
+	local horizontal = element:GetOrientation() == 'HORIZONTAL'
+	local elementSize = horizontal and element:GetWidth() or element:GetHeight()
```
ls-, line 114: I mean this thing, noting it just for ref.

events.lua
```diff
@@ -59,9 +59,9 @@ function Private.UpdateUnits(frame, unit, realUnit)
 			end
 		end
 
-		frame.unit = unit
-		frame.realUnit = realUnit
-		frame.id = unit:match('^.-(%d+)')
+		frame.__unit = unit
+		frame.__realUnit = realUnit
+		frame.__unitIndex = unit:match('^.-(%d+)')
```
ls-: Mb add `tonumber` right here instead of calling it down the line on demand? Index/id being a string feels kinda weird.

ls- (approved):
LGTM 🙏🏼

PR #843, opened 2026-04-01
privateauras: Revert #837

p3lim:
They changed their minds.


ls- (dismissed):
🥷

p3lim:
They added a new option `isContainer` to the anchor API, which is non-optional and will throw errors with it defaulting to `true`, so we have to explicitly set it to `false`. :facepalm:

ls- (approved):
🥀

PR #853, opened 2026-04-22
auras: Revert back to desecreted aura data

p3lim:


ls- (approved):
👌🏻

p3lim:
Gonna hold off on merging this one until I've done some investigating.

p3lim:
Delaying this until 12.0.7 as they're gonna include player vehicle in the `isFromPlayerOrPlayerPet` bool, making it on par with the current filter logic but faster.

elements/auras.lua
```diff
@@ -255,15 +254,15 @@ local function updateAura(element, unit, data, position)
 end
 
 local function FilterAura(element, unit, data)
-	if((element.onlyShowPlayer and data.isPlayerAura) or not element.onlyShowPlayer) then
+	if((element.onlyShowPlayer and data.isFromPlayerOrPlayerPet) or not element.onlyShowPlayer) then
```
Rainrider, line 257: `isFromPlayerOrPlayerPet` is not a replacement for `isPlayerAura`. The former is true if any player or their pet(s) applied the aura, the latter is only for **the** player.

p3lim: `PLAYER` filter which we currently use (`.isPlayerAura` is defined by oUF) covers player, pet and vehicle, while `.isFromPlayerOrPlayerPet` only covers player and pet, but will cover vehicle too in 12.0.7.

Rainrider: `isFromPlayerOrPlayerPet` is ANY player, not THE player.

p3lim: Oh... that's stupid. I'll just close this then.

PR #861, opened 2026-05-16
core: Pass unit to :DisableElement

### LINKED ISSUE #857: unit parameter is always nil in element Disable functions
In alternativepower.lua, the Disable function uses the unit parameter to conditionally restore events on PlayerPowerBarAlt:

```lua
local function Disable(self, unit)
    local element = self.AlternativePower
    if(element) then
        element:Hide()

        self:UnregisterEvent('UNIT_POWER_BAR_SHOW', VisibilityPath)
        self:UnregisterEvent('UNIT_POWER_BAR_HIDE', VisibilityPath)

        if(unit == 'player') then
            PlayerPowerBarAlt:RegisterEvent('UNIT_POWER_BAR_SHOW')
            PlayerPowerBarAlt:RegisterEvent('UNIT_POWER_BAR_HIDE')
            PlayerPowerBarAlt:RegisterEvent('PLAYER_ENTERING_WORLD')
        end
    end
end
```
However, unit is always nil because DisableElement in ouf.lua does not pass it:

```lua
DisableElement = function(self, name)
    -- ...
    return elements[name].disable(self)
end
```
Compare with EnableElement, which does pass the unit:

```lua
EnableElement = function(self, name, unit)
    -- ...
    if(element.enable(self, unit or self.unit)) then
```

As a result, the if(unit == 'player') check in Disable never evaluates to true, and PlayerPowerBarAlt events are never restored when the element is disabled.
Suggested fix:
Either pass self.unit from DisableElement
```lua
return elements[name].disable(self, self.unit)
```
Or use self.unit directly inside Disable:
```lua
local function Disable(self)
    local element = self.AlternativePower
    if(element) then
        element:Hide()

        self:UnregisterEvent('UNIT_POWER_BAR_SHOW', VisibilityPath)
        self:UnregisterEvent('UNIT_POWER_BAR_HIDE', VisibilityPath)

        if(self.unit == 'player') then
            PlayerPowerBarAlt:RegisterEvent('UNIT_POWER_BAR_SHOW')
            PlayerPowerBarAlt:RegisterEvent('UNIT_POWER_BAR_HIDE')
            PlayerPowerBarAlt:RegisterEvent('PLAYER_ENTERING_WORLD')
        end
    end
end
```

p3lim:
Fixes #857

### DIFF
--- a/ouf.lua
+++ b/ouf.lua
@@ -83,7 +83,7 @@ local frame_metatable = {
 Private.frame_metatable = frame_metatable
 
 for k, v in next, {
-	--[[ frame:EnableElement(name, unit)
+	--[[ frame:EnableElement(name[, unit])
 	Used to activate an element for the given unit frame.
 
 	* self - unit frame for which the element should be enabled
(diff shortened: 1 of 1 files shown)


ls- (approved):
🙏🏼

PR #864, opened 2026-06-19
core: Use new rolesets

p3lim:
These are used to conditionally show/hide certain parts of the UI when other UI elements ask for it (i.e. hide unit frames during pet battle), which means we no longer need a separate parent to deal with this.

CUF and others already use it, so if this works as expected we should too.

See [Blizzard_UIModes](https://github.com/BigWigsMods/WoWUI/tree/ptr/AddOns/Blizzard_UIModes) and the [Roleset API](https://github.com/BigWigsMods/WoWUI/blob/ptr/AddOns/Blizzard_APIDocumentationGenerated/RolesetSystemDocumentation.lua)


p3lim:
Got confirmation that this is the intended use-case and should work as expected.

ls- (dismissed):
Finally, the pet battle hider is no more 🙏🏼

ls- (approved):
LGTM 😂

PR #866, opened 2026-06-20
core: Hide non-unit nameplates

### LINKED ISSUE #856: Hide widget-only nameplates and game objects nameplates
**Describe the bug**

oUF creates nameplates for units that should not have custom nameplates displayed, such as game objects and widget-only units. This causes unintended nameplates to appear on fishing bobbers (when using addons like "Better Fishing") and on Hide and Seek game objects. 

In the case of fishing bobbers, the nameplate appears blank or shows stale data from a previously targeted unit. In the Hide and Seek mini-game, these nameplates reveal which objects are players vs. props, giving an unfair advantage.

Image
Image
Image

**How to reproduce**

1. Use any oUF-based nameplate addon.
2. For fishing bobbers: Install the "Better Fishing" addon, cast a fishing line, and observe a nameplate appearing on the bobber.
3. For Hide and Seek: Join a Hide and Seek game and observe nameplates appearing on prop objects.

**What did you expect?**

Nameplates should not be created for units that return `true` for `UnitNameplateShowsWidgetsOnly(unit)` or `UnitIsGameObject(unit)`. These units are not valid targets for custom nameplate rendering.

**Suggested fix**

Add an early return in `driverEventHandler` when handling `NAME_PLATE_UNIT_ADDED`:

```lua
elseif(event == 'NAME_PLATE_UNIT_ADDED') then
    local nameplate = C_NamePlate.GetNamePlateForUnit(unit)
    if(not nameplate) then return end
    if UnitNameplateShowsWidgetsOnly(unit) or UnitIsGameObject(unit) then return end

### PR DESCRIPTION
**Pull request by p3lim**
Fixes #856

### CODE REVIEW

**Comment by p3lim** [comment]
@Paojy please test this one if you can :)

File: ouf.lua [hunk 1]
```diff
@@ -871,49 +874,75 @@ do
...
+
+				if(self.addedCallback) then
+					self.addedCallback(nameplate.unitFrame, event, unit)
 				end
-			end
 
-			if(self.addedCallback) then
-				self.addedCallback(nameplate.unitFrame, event, unit)
-			end
+				if(isUnit) then
```
ls-: Why is this check even here? This whole block is behind the `isUnit` check.

ouf.lua
```diff
@@ -871,49 +874,75 @@ do
...
+
+					-- enable elements if they were previously disabled
+					if(previouslyActiveElements[nameplate.unitFrame]) then
+						for element in next, previouslyActiveElements[nameplate.unitFrame] do
+							nameplate.unitFrame:EnableElement(element, unit)
+						end
 
-			nameplate.unitFrame:SetAttribute('unit', unit)
+						table.wipe(previouslyActiveElements[nameplate.unitFrame])
+						previouslyActiveElements[nameplate.unitFrame] = nil
```
ls-: Shouldn't `wipe` alone suffice?

ouf.lua
```diff
@@ -871,49 +874,75 @@ do
...
 
-			nameplate:ClearAllHitTestPoints() -- to prevent lingering hit test points on default
-			nameplate:SetAllHitTestPoints(nameplate.unitFrame)
+					if(nameplate.UnitFrame.SoftTargetFrame) then
+						nameplate.UnitFrame.SoftTargetFrame:SetParent(nameplate)
+					end
+				else
+					Private.UpdateUnits(nameplate.unitFrame, unit)
+
+					-- enable elements if they were previously disabled
```
ls-: Honestly, maybe turn this whole element juggling into something like Pause/ResumeElements methods for all of our unit frames? Idk.

ouf.lua
```diff
@@ -871,49 +874,75 @@ do
...
+					-- ForceUpdate calls layouts have to do after changing things
+					nameplate.unitFrame:UpdateAllElements(event)
+				end
+			elseif(nameplate.unitFrame) then
+				previouslyActiveElements[nameplate.unitFrame] = {}
 
-			-- UAE is called after the callback to reduce the number of
-			-- ForceUpdate calls layouts have to do after changing things
-			nameplate.unitFrame:UpdateAllElements(event)
+				for element in next, activeElements[nameplate.unitFrame] do
```
ls-: See https://github.com/oUF-wow/oUF/pull/866/changes#r3468499642

ls- (changes requested):
🤔

ouf.lua
p3lim: :D yeah I've already refactored this away locally

ouf.lua
p3lim: Yeah it should

ouf.lua
p3lim: Hm, we could, but how'd you go about doing that?

ls- (approved):
Yeah, I go ahead. I tinkered with my own suggestion at some point, but the idea of having a method that's basically only useful for nameplates on all unit frames seemed questionable. I mean that pause/resume elements thingy we talked about internally. I just don't see it being useful elsewhere.

p3lim:
We might have to revisit that for rolesets, but for now I'll merge this in.

PR #863, opened 2026-06-17
castbar: handle nil in UpdatePips

KogasaPls:
The Blizzard docs suggest this should never be nil, but I managed to get a lua error implying it was anyway. Handle this safely by replacing nil with an empty table.

PR #867, opened 2026-06-21
castbar: Reset before we start a cast

p3lim:
Unsure what could cause this, maybe a layout messed around with the castbar attributes after it was reset, so we'll just do this to be safe.

Should fix #863


ls- (approved):
👌🏼

PR #868, opened 2026-06-21
pingindicator: New element

p3lim:
See UnitPingIconFrameMixin and UnitPingIconFrameTemplate in Blizzard_PingUI for reference.

### DIFF
--- a/elements/pingindicator.lua
+++ b/elements/pingindicator.lua
@@ -1,0 +1,121 @@
+--[[
+# Element: PingIndicator
+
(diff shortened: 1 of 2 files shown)


p3lim:
Currently this element is nonoperational as the events are restricted (i.e. we can't register them), Blizzard is looking into relaxing that if possible.

ls- (approved):
🥳

p3lim:
We won't be able to do this in 12.1, Blizzard has stated that we could possible see this functionality in a later patch.

PR #870, opened 2026-07-14
auras: Rewrite for 12.1

p3lim:
In broad terms, the way we've been doing auras for years in oUF is now over, we hand all of the logic over to Blizzard and we just wrap their new shiny way of dealing with auras in a wrapper that makes it a bit more familiar to users of oUF, and deal with some of the necessary manual handling of unit updates and such so the layouts don't have to (very much on-brand for oUF).

Don't look at this PR as a diff, pretend like the original auras implementation is now fully gone and think of the new one as a new implementation, because that's what is really is - they share nothing alike but some option names and the button template (to some degree). I.e. just look at the file directly: https://github.com/oUF-wow/oUF/blob/aurapocalypse/elements/auras.lua

The element is now invoked similar to tags, as layouts request aura containers through a new meta method on each unit frame, with some additional methods attached to make it slightly easier to work with (although they're not necessary to use if you don't want to - i.e. if you replaced `CreateButton` before then you don't need to use the extra methods at all).

### DIFF
--- a/elements/auras.lua
+++ b/elements/auras.lua
@@ -1,8 +1,8 @@
 --[[
 # Element: Auras
 
-Handles creation of [aura containers](https://warcraft.wiki.gg/wiki/UIOBJECT_AuraContainer), groups,
-slots, and [buttons](https://warcraft.wiki.gg/wiki/UIOBJECT_AuraButton).
+Handles creation of [aura containers](https://warcraft.wiki.gg/wiki/INTRINSIC_AuraContainer), groups,
+slots, and [buttons](https://warcraft.wiki.gg/wiki/INTRINSIC_AuraButton).
 
 ## Notes
 
@@ -209,8 +209,8 @@ local elementMixin = {}
 Defines a group of auras to display on the element.  
 This can be defined multiple times.
 
-* filter  - aura filter for this group ([AuraFilter](https://warcraft.wiki.gg/wiki/API_type/AuraFilters))
-* options - options for this group (TODO: link to wiki)
+* filter  - aura filter for this group ([AuraFilter](https://warcraft.wiki.gg/wiki/API_types/AuraFilters))
+* options - options for this group ([CustomAuraContainerGroupDefaultOptions](https://warcraft.wiki.gg/wiki/FrameXML_types/CustomAuraContainerGroupDefaultOptions))
 
 ## Notes
 
@@ -279,8 +279,8 @@ Defines a slot for a single buff or debuff to create from the element.
 The slot can be manually positioned if necessary.  
 This can be defined multiple times.
 
-* filter  - aura filter for this group ([AuraFilter](https://warcraft.wiki.gg/wiki/API_type/AuraFilters))
-* options - options for this group (TODO: link to wiki)
+* filter  - aura filter for this group ([AuraFilter](https://warcraft.wiki.gg/wiki/API_types/AuraFilters))
+* options - options for this group ([CustomAuraContainerSlotDefaultOptions](https://warcraft.wiki.gg/wiki/FrameXML_types/CustomAuraContainerSlotDefaultOptions))
 
 ## Notes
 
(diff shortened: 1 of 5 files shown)


p3lim:
There's one really ugly part of this PR; AuraContainer frames cannot be created during combat, which is when nameplates and group header children could potentially spawn, so we need to pre-allocate them. Since layouts are unpredictable we'll need to expose an API to let them adjust the number of containers on the fly. The defaults are still TBD.

Also, since this is no longer a "real" element (like tags) it's necessary to add a little bit of jank in order for UAE to work.

Lastly, this has not been tested throughly since I've been sick for the past week, and will continue to be in a limbo until I get better, so please, if you're reading this, test!

p3lim:
Updated my personal layout(s) to use the new element, with backwards-compatibility: https://github.com/p3lim-wow/Inomena/commit/8c693431fb10235f74a8cef4f494993e44cb5f5b

I guess this could be seen as an example of how one could migrate over to the new element.

p3lim:
Next build will fix the reset issue and we can drop most of our state, and the hacks mentioned in https://github.com/oUF-wow/oUF/pull/870/changes/3c804d7d6314192b5cf10c4c7db03918f0376a28.

Once that has been fixed I'll rebase the branch and get ready for merging.

elements/auras.lua
```diff
@@ -1,938 +1,471 @@
...
+.layoutLimit   - Max width or height of the element, depending on the layout. Defaults to the parent width or height (number?)
+.initialAnchor - Anchor point for the element. Defaults to 'TOPLEFT' (string?)
+.growthX       - Horizontal growth direction. Defaults to 'RIGHT' (string?)
+.growthY       - Vertical growth direction. Defaults to 'UP' (string?)
+.padding       - Padding around the element. Defaults to 0 (number?)
+.paddingLeft   - Padding on the left side of the element. Takes priority over `padding` (number?)
+.paddingRight  - Padding on the right side of the element. Takes priority over `padding` (number?)
+.paddingTop    - Padding on the top side of the element. Takes priority over `padding` (number?)
+.paddingBottom - Padding on the bottom side of the element. Takes priority over `padding` (number?)
+.policies      - Policy for how auras should be processed. See CustomAuraContainerProcessAuraPolicyDefaultOptions (table?)
```
ls-: A bit nitpicky, but mb rename it to `policy`? It's a policy, it goes to `SetAuraProcessingPolicy` and whatnot.

p3lim: Fair, I'll rename it.

ouf.lua
```diff
@@ -1013,6 +1013,32 @@ function oUF:AddElement(name, update, enable, disable)
...
+* self    - the global oUF object
+* name    - unique name of the element (string)
+* create  - used to create the element. Will be registered as a meta function (function)
+* update  - used to update the element (function)
+* enable  - used to enable the element for a given unit frame and unit (function)
+* disable - used to disable the element for a given unit frame (function)
+--]]
+function oUF:AddMetaElement(name, create, update, enable, disable)
+	argcheck(name, 2, 'string')
+	argcheck(update, 3, 'function')
```
Rainrider: Do you mean `create` here?

p3lim: Good catch :+1:

elements/auras.lua
```diff
@@ -1,938 +1,444 @@
...
+		local paddingLeft = options.paddingLeft or options.padding or 0
+		local paddingRight = options.paddingRight or options.padding or 0
+		local paddingTop = options.paddingTop or options.padding or 0
+		local paddingBottom = options.paddingBottom or options.padding or 0
+		element:SetFlowLayoutPadding(paddingLeft, paddingRight, paddingTop, paddingBottom)
 
-			if(debuffs.PostUpdate) then debuffs:PostUpdate(unit) end
+		if(options.policy) then
+			-- just expose it easily for layouts in case they want to use it
+			element:SetAuraProcessingPolicy(CustomAuraContainerAuraProcessingPolicy.ProcessAura, options.policy  )
```
ls-: You fixed extra whitespaces above, but forgot this one 😅

ls- (dismissed):
🤯

ls- (dismissed):
Bruh 😒

elements/auras.lua
```diff
@@ -414,8 +414,8 @@ end
 local function Update(self)
 	if(STATE[self] and STATE[self].elements) then
 		for _, element in next, STATE[self].elements do
-			if element:GetUnit() ~= self.unit then
-				element:SetUnit(self.unit) -- triggers a full update
+			if element:GetUnit() ~= self.__unit then
+				element:SetUnit(self.__unit) -- triggers a full update
```
ls-, line 418: Double checked jic, there are no other `.unit`s anywhere.

ls- (approved):
BRUH 🙄

PR #881, opened 2026-08-23
health/totem: Ensure widgets are shown

p3lim:
After calling Disable (perhaps through Pause) these widgets will never be shown again (from Enable or Resume), this fixes that.


ls- (approved):
😒

PR #882, opened 2026-08-23
raidroleindicator: Account for secret assigments

p3lim:
While #880 accounts for secret assignments it would hide the indicators outright if the assignments were secret, removing existing information from the player. This PR will use a cached state while secrets are in effect, and do an update again when secrecy is (presumably) gone after combat ends.

Closes #880

### DIFF
--- a/elements/raidroleindicator.lua
+++ b/elements/raidroleindicator.lua
@@ -29,6 +29,8 @@ This element updates by changing the texture.
 local _, ns = ...
 local oUF = ns.oUF
 
+local STATE = {}
+
 local function Update(self, event)
 	local element = self.RaidRoleIndicator
 	local unit = self.__unit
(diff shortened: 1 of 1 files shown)


ls- (approved):
While I'm not particularly thrilled about cached info, it's better than no info, I guess 😩

ls- (changes requested):
Bruh...

ls- (approved):
Soz, was tripping balls, it's fine.

ls- (changes requested):
Ugh, basically, my concern is that what if an old unit leaves the raid, but in their place another one joins, then the new unit will be using cached info from the previous unit.

p3lim:
If that happens while secrets are in effect then that new unit will have the state of the old one, correct, but after combat drops or if the old information was from a secret then the info gets a fresh update for the new unit.

ls-:
Yeah, ik that, but it doesn't make it any less annoying. Mb purge STATE when `OnShow` fires? But then there's the UI hiding bs like alt-z or cutscenes, but `Private.UpdateUnits` should take care of that...

p3lim:
I'll test that.

p3lim:
It will be affected by alt-z and cutscenes indeed, but realistically noone uses alt-z in the middle of a real raid fight, and cutscenes are always(?) at the end of fights if there is one, so I don't think either will be a real problem.

ls- (approved):
🙏🏼

PR #883, opened 2026-08-23
pvpindicator: Use cached PvP state when secret

p3lim:
Similar to #882 this uses cached PvP state while secrets are in effect.

We should consider a way to reset this based on some unit identifier, but the whole point of this secrecy is that we can't identify units so it's probably not possible.


ls- (approved):
Likewise 😩

ls- (changes requested):
Bruh...

ls- (approved):
Ugh, whatevs.

ls- (changes requested):
Yeah, nah for now. 

Image

ls- (approved):
🙏🏼

PR #884, opened 2026-08-23
health/power: Normalize floating point errors

p3lim:
Would like to take scale into account here, especially for nameplates, but nameplate scaling is weird so I'll ignore it for now.


ls- (approved):
🙏🏼

Issue #888, opened 2026-08-31
privateauras passes showCountdownFrame, removed from AddPrivateAuraAnchorArgs in 12.1


Booyaka101:
**Describe the bug**

`elements/privateauras.lua` line 133 passes `showCountdownFrame`:

```lua
showCountdownFrame = not element.disableCooldown,
showCountdownNumbers = not element.disableCooldownText,
```

In 12.1 `AddPrivateAuraAnchorArgs` has no `showCountdownFrame`. It is `showCooldownFrame` now, and the structure also gained `showCooldownEdge` and `showDispelIcon`:

```lua
{ Name = "showCooldownFrame", Type = "bool", Nilable = false, Default = false },
{ Name = "showCooldownEdge", Type = "bool", Nilable = false, Default = false },
{ Name = "showCountdownNumbers", Type = "bool", Nilable = false, Default = false },
{ Name = "showDispelIcon", Type = "bool", Nilable = false, Default = false },
```

`showCountdownNumbers` on line 134 was not renamed, so only the first of the pair is affected. `showCooldownFrame` never gets set and its documented default is false, so the spiral stays off and `.disableCooldown` on line 14 has nothing to switch either way.

Blizzard's own callers use the new names, `Blizzard_UnitFrame/Shared/CompactUnitFrame.lua` and `Blizzard_BuffFrame/BuffFrame.lua` both pass `showCooldownFrame` next to `showCountdownNumbers`.

KkthnxUI vendors this element and made the same rename in Kkthnx-Wow/KkthnxUI@8522b82 on 28 Aug. SpartanUI's copy still has the old name.

**How to reproduce**

1. Create a PrivateAuras element per the docs and leave `.disableCooldown` unset
2. Take a private aura
3. The icon and the duration text appear, the cooldown spiral does not

**What did you expect?**

The spiral to draw, and `.disableCooldown = true` to be the thing that turns it off.

**Error report**

```
None.
```

I read this off the 12.1 structure docs and Blizzard's callers rather than running it, and I have not tested whether the stale key is ignored or rejected, so it is worth a sanity check before you take it. The fix is the rename on line 133.

Found it with a linter I have been writing for the 12.1 secret value changes, https://github.com/Booyaka101/wow-secret-lint

p3lim:
[...]

PR #889, opened 2026-08-31
privateauras: Remove element

p3lim:
The PrivateAuras element is hereby deprecated.

Ref #888


ls- (approved):
I honestly thought it's already done 😅

PR #893, opened 2026-09-16
pvpindicator: Use new Util, add new sub-widget

### LINKED ISSUE #890: Midnight: can't display pvpindicator in pvp content
UnitHonorLevel and UnitIsPVP both return secrets in PvP instances, which makes the pvpindicator element non-functional.

Blizzard said they'll look into solutions for us.

Ref #886 & #887.

p3lim:
Fixes #890

This also removes the `status` payload from the callback, which would technically force a new major version, but we'll ignore it in this case since it's currently broken for many unit types.


p3lim:
Will rebase this and use new conditional stuff from #898 so we can release faster.

p3lim:
Gonna revise this a bit, but can't test right now since PTR is dead.

PR #896, opened 2026-09-20
power: fix power type coloring

Rainrider:
This fixes power type coloring so that it respects the intended color priority:

1. by power token
2. by power alt colors as provided by UnitPowerType
3. by power type

Previously it was using the power type when there was a color matched by power token.

This can be tested with the [Tending the Garden](https://www.wowhead.com/quest=28751/tending-the-garden) quest, where `UnitPowerType('vehicle')` returns

```
[1]=3,
[2]="POWER_TYPE_SUN_POWER"
[3]=1,
[4]=0,
[5]=0,
```


p3lim:
Yeah this is the same logic the tag is using.

p3lim (approved):

Issue #895, opened 2026-09-16
Support "Forever" game version


p3lim:
The game will be based on 12.1.5 and will be in the mainline branch (same as retail), so we should be able to support it right out of the box. Will have to test once the beta is out.

Stagger and Runes already handle the APIs not existing (they disable themselves if the player is not the correct class), but we'll need to check all the elements to be sure.

The intent here is to automatically disable elements that are not suitable for that game version, so layouts won't need to do too many game version checks themselves to show/hide elements. The target for the support would be the same as 12.1.5, so we might need to do compatibility checks internally depending on when these two releases ship, in order to be able to release for one without ruining the other.

Some things that come to mind that we'd want to support that we currently lack:

- Pet happiness:
  - color on health bar
  - tag
  - element for texture widget
- master looter element
- ...?

Known issues:

- restricted environment is currently broken, so headers can't be spawned
- combo points API is currently secret

arnvid:
Two things from getting RealUI running on the beta:

**Arena frames:** `CompactArenaFrame` doesn't exist on Forever, so spawning any `arena` unit errors in `blizzard.lua` (L119, the `memberUnitFrames` loop). Blizzard guards it themselves in `Blizzard_CompactRaidFrameContainer.lua` (`if CompactArenaFrame and ...`). `ArenaEnemyMatchFramesContainer` only comes from `Blizzard_Deprecated_ArenaUI` there, so it probably wants the same guard.

**Restricted environment:** tracked it down, it's a TOC load-order thing on Blizzard's end. `Blizzard_EnvironmentCleanup` is `LoadFirst: 1` and nils `loadstring_untainted`, but its dependency line is

```
## Dep: Blizzard_RestrictedAddOnEnvironment [AllowLoadGameType classic, standard]
```

with no `camelot`. So on Forever the cleanup can run before `RestrictedExecution.lua:22` grabs the local, and every snippet then dies at L79 with "attempt to call a nil value". Nothing we can do addon-side beyond skipping headers until they add `camelot` to that line. But I think you know about these ones.

PR #898, opened 2026-09-21
private: Tables to check game version or compat

p3lim:
Will be heavily used for game version compatibility for Forever (see #895), but also for PTR/beta, making it easier for us to release versions and keep progress in the master branch, so layout authors can get in on testing upcoming stuff quicker.


ls- (approved):
Go go go!

PR #899, opened 2026-09-22
units: Fix arenaprep

p3lim:
We never migrated this to use the element pause logic in v14, and it referenced pre-v14 elements like auras. This should fix that, but it needs testing before it can be merged.


p3lim:
This is for 12.1.0 and should be fine to review and merge. Going to do a follow-up PR for 12.1.5 afterwards, as it added a few new arena-related changes.

PR #903, opened 2026-09-22
happiness: Resurrect element from the past

p3lim:
This is pretty much identical to the element we killed off in 13e9782.

Ref #895


p3lim:
Note: there is a TODO comment at the bottom of the happiness element, should consider making it a frame (or attach one) to display a tooltip with more information, as we do with other elements (like phaseindicator).

PR #904, opened 2026-09-22
masterlooterindicator: Resurrect element from the past

p3lim:
Pretty much identical to the element we killed off in 32aa62e.

Ref #895


p3lim:
Note: there's a comment about GROUP_ROSTER_UPDATE in Enable/Disable, need to check that before we merge.

PR #908, opened 2026-09-22
core: Add GetUnitFrame method

p3lim:
Useful for integrating something with oUF.


arnvid:
since 14 dropped `frame.unit` there's no public way to get a frame's live unit (the vehicle-aware one), and Blizzard's `UnitFrame_OnEnter` still reads `self.unit`, so the usual `SetScript('OnEnter', UnitFrame_OnEnter)` gives no tooltip. We're copying `__unit` into `.unit` on hover for now, which obviously isn't great. Would a `frame:GetUnit()` (or keeping `.unit` around read-only) be on the table?

